PR #1, opened 2019-11-30
Fix sitting in net/minecraft

Chocohead:
Technically I do have a Loom fork with access transformers too, but in this case you can work around the problem from Mojang kindly making `IdentityListMerger` as basically as pure as you could possible want to override. That solves the problem for `IndirectListPair` as it can just extend it and override its only two methods (the two `DoubleListPair` needs).

The other overrides for `VoxelShapes` needed a bit more creativity to work around having explicit types but with some injecting and redirecting it solves needing to do anything more dramatic. Don't think anything which is done is any worse than what was there before in the grand scheme of things.


jellysquid3:
Thanks for taking a look into this. I think the solution you're proposing here is far better than mucking about with packages and looks good to me.

PR #3, opened 2019-12-04
Add Gradle wrapper

magneticflux-:
Checked out `master` and had to reinstall and guess the version of Gradle needed to build it.

I assumed the latest 5.x version for compatibility.

The execute bit is set on `gradlew` so it should be good to go for non-Windows.


jellysquid3:
The README file specifies the preferred version of Gradle [here](https://github.com/jellysquid3/Lithium#compiling-the-mod), which is Gradle 5.5. I've had issues with newer versions, though I'm not sure if that's because of the buildscript or just tooling issues. It seems to work fine in my testing right now... leading me to believe that the issues were probably related to the now-removed shadow plugin.

Regardless, the missing wrapper is an oversight. I use SDKMAN on Fedora to manage Gradle installations so the wrapper sees little usage on my system, but I'm happy to add it for users who prefer it. Thanks for the PR.

magneticflux-:
I use SDKMAN as well, but I prefer to use the wrapper for this very reason - when a project needs an older Gradle version to build.

Glad I could help!

PR #4, opened 2019-12-06
Items now insert themselves into hoppers rather than hoppers searching the entity list

Ueaj-Kerman:
The option is disabled by default because of an edge case pointed out by 2no, which is documented under the config option. There may be a slightly better way which is also documented, but I am unsure how to implement it correctly and if it has other side effects.


Devan-Kerman:
There is a problem, if the hopper is full while an item enters the range, it'll get stuck inside the bowl, will fix now

Devan-Kerman:
I forgot to update the cooldown for hoppers :p and I fixed the other thing too

Devan-Kerman:
Hm I found another bug, for some odd reason, the hopper's block state wont update properly

Devan-Kerman:
I've tested item filters and a few other things to make sure most vanilla things work and they do now

liach:
Is this worth it? imo usually players don't build that many hoppers but there is way more drops

src/main/java/me/jellysquid/mods/lithium/mixin/entity/item_self_insert/HopperBlockEntityMixin.java
```diff
@@ -0,0 +1,72 @@
...
+    @Shadow
+    private int transferCooldown;
+
+    public HopperBlockEntityMixin(BlockEntityType<?> type) {
+        super(type);
+    }
+
+    @Inject(method = "extract(Lnet/minecraft/block/entity/Hopper;)Z", cancellable = true, at = @At(value = "INVOKE", target = "Lnet/minecraft/block/entity/HopperBlockEntity;getInputItemEntities(Lnet/minecraft/block/entity/Hopper;)Ljava/util/List;"))
+    private static void extract(Hopper hopper, CallbackInfoReturnable<Boolean> cir) {
+        if (hopper instanceof HopperBlockEntity && !((HopperBlockEntityMixin) hopper).full()) {
```
liach: This cast shall fail as the `HopperBlockEntity` won't be an instance of `HopperBlockEntityMixin` at runtime. Consider exposing the method to `HopperAccess` instead

liach (changes requested):

asiekierka:
I'd say it's a good idea in a more vanilla environment, but may not be so useful in a modded environment.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/item_self_insert/HopperBlockEntityMixin.java
Devan-Kerman: Oops

Devan-Kerman:
In my tests it has improved performance, it's either each hopper searching all nearby entities or each entity searching for a few hoppers.

jellysquid3:
I've been thinking about the implications this PR would have for the project over the last few days. In particular, the primary question is whether or not we should change vanilla behaviour in the goal of better performance. 

This patch obviously does has a significant impact on performance in these rarer cases involving many hoppers but comes with a serious caveat: It can change certain expected behaviour in vanilla Redstone contraptions and changes semantics slightly for item entities.

I don't think Lithium should be the project to be making these kinds of changes. It's my number one goal for this mod to act in a completely transparent manner and I think that point alone has a drastic impact on how people view the work being done here.

However, that is not to say that I'm completely uninterested in this patch or others like it. At some point in the nearby future, I would like to create a companion mod for Lithium which permits these sorts of changes in the name of raw performance. But right now my priorities remain to be further improvement on Lithium and Phosphor.

As such, I'm going to be putting this pull request on the back-burner for a little bit while things settle. Thank you for your work so-far on this patch set, @Devan-Kerman.

Devan-Kerman:
alright cool, will you use the new entity chunk cache or should I make another pr for that

jellysquid3:
It'd be great if you could open a PR for the entity chunk cache improvements on a different branch so we can track the two separately. We've discussed the changes on Discord already and I'd like to look at incorporating the specific patch. :smile:

src/main/java/me/jellysquid/mods/lithium/mixin/entity/simple_entity_block_collisions/MixinViewableWorld.java
```diff
@@ -42,10 +42,10 @@
      */
     @Overwrite
     default boolean doesNotCollide(Entity entity, Box box, Set<Entity> ignore) {
-        EntityChunkCache entityChunkCache = null;
+        EntityChunkCache EntityChunkCache = null;
```
natanfudge: The capitalized E here is probably not intentional

2No2Name:
I just saw this by chance, also devan mentioned me for some reason. I have an implementation for this and it works quite well. Since items are already checking for collisions anyways, I am using that getBlockState to find hoppers and then put them on an entity list that each hopper is keeping. The update order doesn't change, only the order in which items may be picked up changes in edge cases (https://github.com/2No2Name/hopperOptimizations/issues/1). I also need hoppers doing checks in lazy chunks and when they are first ticked. Same functionality is used for storage minecarts. I don't think it is needed for you to reimplement this, as long as our mods are compatible

jellysquid3:
Looks like GitHub closed this issue and won't re-open it after I moved some branches around.

PR #5, opened 2019-12-12
entity chunk cache 2: electric boogaloo

Devan-Kerman:
a faster entity chunk cache


jellysquid3:
Thank you for the contribution. In particular, re-using the chunks from the cache when updating it is a sneaky trick which helps a bit. :) 

I've modified your implementation to be slightly clearer and to address some other issues with the entity chunk cache (in particular, we need to invalidate data if an entity is moved between worlds.)

PR #6, opened 2019-12-17
Refactor world border collision

silversquirl:
This PR moves the optimized world border collision code into the `LithiumEntityCollisions` class, which allows it to be used in the general entity collision optimizations.


vktec:
Probably a good idea to squash commits; I pushed too early

liach (approved):

jellysquid3 (changes requested):
These changes break collisions against the world border while the player or another entity is outside of it, making it impossible to cross the planes the world border starts at. To replicate the issue, create a small world border and place the player outside of it by no-clipping with F3+N. Then, disable no-clip and try to move around one corner of the world border (while still outside of it.)

jellysquid3:
Apologies. The issues I've detailed above are not a result of your implementation and are instead rather edge cases in the original implementation that only occurred in some situations. I've since fixed it and have replaced the calls to `VoxelShapes#matchesAnywhere(...)` in `getBlockCollisions` with our own.

PR #23, opened 2020-03-26
ChunkRandom optimizations

KaptainWutax:


2No2Name:
I assume that your code is taken from a different repository, so make sure to agree with CONTRIBUTING.md . Have you done any testing whether there is a performance improvement and whether the code works correctly/does not deviate from vanilla? Does the performance boost only occur in the no longer O(n) consume(...) or also the other random calls?
The code looks good to me. I noticed that vanilla only uses 262,17292 and 2620 as the argument of consume(...). So recalculating the LCG for these values is probably not necessary, instead a small cache could be used. What do you think?

KaptainWutax:
Yes, this has been thoroughly tested and provides around 6-7 times speedup on rng calls. Will create a new PR soon so I'm closing this one.

PR #25, opened 2020-03-29
Update for 20w13b

smonnnn:
I've been running 20w13b with these changes to lithium for 3 days and it has been working without issue so far. Would be nice to have some more confirmation though.


jellysquid3:
Thanks for the contribution! I had most of the port already present in my local source tree -- problem was fixing up a few names in Redstone code and pushing. I'll be closing this as f3430ab69845797b8c7a94efb6936b9d160306c8 has landed.

smonnnn:
Nice, thanks!

PR #32, opened 2020-04-14
change: Improved performance of Voronoi a bit more.

Kroppeb:
Changes the implementation of `distribute`

One further optimization that could be possible is to check all 8 biomes, in case they are equal just return that one.

JMH:
```
Benchmark                   Mode  Cnt          Score          Error  Units
MyBenchmark.newDistribute  thrpt   25  300172640.366 ± 10675013.061  ops/s
MyBenchmark.newGetBiome    thrpt   25    6996474.516 ±    89029.921  ops/s
MyBenchmark.oldDistribute  thrpt   25  185714728.519 ±  3260407.310  ops/s
MyBenchmark.oldGetBiome    thrpt   25    5279638.474 ±   352525.128  ops/s
```

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/biome/MixinVoronoiBiomeAccessType.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/biome/MixinVoronoiBiomeAccessType.java
@@ -85,12 +85,12 @@ public abstract class MixinVoronoiBiomeAccessType {
     }
 
     /**
-     * @reason Replace expensive modulo with simple bit shift
-     * @author JellySquid
+     * @reason Replace expensive modulo with simple `bitwise and` and reduced flops
+     * @author Kroppeb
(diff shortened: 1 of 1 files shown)


Kroppeb:
I also did I check for all values between `0` and `1024` (exclusive) bit-shifted to the left `24` and used that as `seed` to verify it still returns the same values.

jellysquid3:
Wouldn't the change from a floating point division to a inverse multiplication introduce accuracy/rounding errors and change the result of the function?

jellysquid3:
It appears not! I'm able to successfully reproduce your results -- looks good to me for merging.

Kroppeb:
> Wouldn't the change from a floating point division to a inverse multiplication introduce accuracy/rounding errors and change the result of the function?

No as `1024` is a power of 2 so it shouldn't affect the mantissa.

PR #34, opened 2020-04-15
Entity getting optimizations

2No2Name:
New optimizations in this pullrequest:

- Add getting only boats and shulkers when getting collision boxes (collision_only_boat_shulker)
    - will lead to problems with mods that add entities with a collision box like boats / shulkers
    - contains a workaround to be able to group entities by an interface in typefilterablelist
- Add type specific entity retrieval to replace predicates that filter entity types (replace_entitytype_predicates)
- Reorder entity to block and entity to entity collisions to query entities only if necessary (stream_entity_collisions_lazily)
    - entities standing on the ground no longer search for entities (boat/shulker) to collide with when trying to move downwards due to gravity


src/main/java/me/jellysquid/mods/lithium/common/entity/LithiumEntityCollisions.java
```diff
@@ -251,4 +258,29 @@ public static boolean isBoxFullyWithinWorldBorder(WorldBorder border, Box box) {
...
+
+    /**
+     * Partial [VanillaCopy] Classes overriding Entity.getHardCollisionBox(Entity other) or Entity.getCollisionBox()
+     * @param entityView the world
+     * @param selection the box the entities have to collide with
+     * @param entity the entity that is searching for the colliding entities
+     * @return list of entities with collision boxes
+     */
+    public static List<Entity> getEntitiesWithCollisionBoxForEntity(EntityView entityView, Box selection, Entity entity) {
+        if (entity instanceof BoatEntity || entity instanceof StriderEntity || entity instanceof MinecartEntity || !(entityView instanceof World)) {
```
jellysquid3: `collision_only_boat_shulker` looks particularly fragile here. I don't feel comfortable adding it on the basis that it relies on having assumed knowledge on what entities may or may not do in the game.

src/main/java/me/jellysquid/mods/lithium/common/world/ExtendedChunk.java
```diff
@@ -0,0 +1,11 @@
+package me.jellysquid.mods.lithium.common.world;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.util.math.Box;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public interface ExtendedChunk {
+    <T> void getEntitiesCustomType(Entity excluded, Class<? extends T> class_1, Box box_1, List<Entity> list_1, Predicate<? super T> predicate_1);
```
jellysquid3: The method and interface names here is are rather vague. Maybe `CustomFilterableEntityList` would be better and the `ExtendedWorld2` interface could be merged into this one?

src/main/java/me/jellysquid/mods/lithium/mixin/collections/entity_filtering/MixinTypeFilterableList.java
```diff
@@ -43,7 +44,7 @@
     }
 
     private <S> Collection<T> createAllOfType(Class<S> type) {
-        if (!this.elementType.isAssignableFrom(type)) {
+        if (type != LithiumEntityCollisions.CollisionBoxOverridingEntity.class && !this.elementType.isAssignableFrom(type)) {
```
jellysquid3: The reason for this specific check is not clear and should be documented.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/collision_only_boat_shulker/MixinBoatEntity.java
```diff
@@ -0,0 +1,8 @@
+package me.jellysquid.mods.lithium.mixin.entity.collision_only_boat_shulker;
+
+import me.jellysquid.mods.lithium.common.entity.LithiumEntityCollisions;
+import net.minecraft.entity.vehicle.BoatEntity;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(BoatEntity.class)
+public class MixinBoatEntity implements LithiumEntityCollisions.CollisionBoxOverridingEntity { }
```
jellysquid3: Instead of adding marker types to entity classes through Mixins, we should prefer to build a registry in Lithium (a simple HashMap linking EntityType to a special enum would be sufficient). Ultimately we want to avoid Mixins as much as possible and keep it maintainable.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/collision_only_boat_shulker/MixinShulkerEntity.java
```diff
@@ -0,0 +1,8 @@
+package me.jellysquid.mods.lithium.mixin.entity.collision_only_boat_shulker;
+
+import me.jellysquid.mods.lithium.common.entity.LithiumEntityCollisions;
+import net.minecraft.entity.mob.ShulkerEntity;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(ShulkerEntity.class)
+public class MixinShulkerEntity implements LithiumEntityCollisions.CollisionBoxOverridingEntity { }
```
jellysquid3: See comment about a marker type registry.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/collision_only_boat_shulker/MixinWorld.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Shadow
+    public abstract Profiler getProfiler();
+
+    @Shadow
+    public abstract ChunkManager getChunkManager();
+
+    //[VanillaCopy] custom: class filtered entity list together with excluding one entity
+    @Override
+    public <T> List<Entity> getEntitiesCustomType(Entity excluded, Class<? extends T> class_1, Box box_1, Predicate<? super T> predicate_1) {
```
jellysquid3: It doesn't appear that this needs to be done in a Mixin and could instead be moved to a static helper class.

jellysquid3 (changes requested):
This PR contains a number of obfuscated names still and should be cleaned up.

src/main/java/me/jellysquid/mods/lithium/common/entity/LithiumEntityCollisions.java
2No2Name: I agree it is very fragile with other mods, the alternative would be to check for each entity type whether it overrides getCollisionBox or getHardCollisionBox at runtime using reflection and cache the result.

2No2Name:
I am splitting the pullrequest up in smaller parts, will reopen new ones soon

PR #38, opened 2020-05-04
Fix Creepers sometimes not pathfinding to the player caused by incorrect chunk cache initialization

2No2Name:
Lithiums flattened chunk cache is initialized with the wrong size (too small by 1 or 2 chunks in both directions). Therefore important chunks for mob navigation are missing in the cache, and assumed to be air, which will make the mob not walk like in vanilla. On discord there was a report of one creeper not tracking the player, I reproduced this and the corrected chunk cache size is fixing it.


jellysquid3:
I tipped off another user that this was the likely culprit, never got around to looking into the exact source of the issue. Good find.

PR #41, opened 2020-05-11
Allow hotswapping by removing "switch on enum" in mixin classes.

2No2Name:
Hotswapping failed probably due to the class required for switch on enum: https://www.benf.org/other/cfr/switch-on-enum.html

After following the tutorial at https://fabricmc.net/wiki/tutorial:mixin_hotswaps . I still couldn't hotswap mixins in lithium, while it works perfectly fine for other mods. From my testing it looks like that using switch on enum in mixin classes breaks hotswap. After replacing the switch on enum with switching on the ordinal integer hotswap is working again.


jellysquid3:
Admittedly, this code is not as pleasant as the switch-on-enum syntax, but it's evident that this generates some kind of craziness in the bytecode and that it's responsible for breaking hotswap functionality.

It's worth measuring again to see if these patches are even still relevant. They were written before we had shape specialization and the JVM had difficulty optimizing/inlining the calls in VoxelShape.

Earthcomputer:
This has been affecting me to, glad someone found the cause. I reported it to the Mixin repository: https://github.com/SpongePowered/Mixin/issues/423

PR #42, opened 2020-05-13
Evaluate getEntityCollisions lazily and place block collisions earlier in the stream

2No2Name:
Getting entities from the world is a hotspot in movement collision code. As most movements are just caused by gravity and are cancelled as soon as the collision with the block below the entity is checked, I suggest not getting entity collision shapes from the world at all, if blocks already stop the movement completely.
For this two things have to be done:
- In the collision voxelshape stream the block collisions have to be earlier than the entity collisions.
- If the entity collisions in the stream aren't used, don't get the entities from the world.

The 2nd one seems to be a bit tricky, but one solution is to use a supplier lamda that captures the arguments for the getEntityCollisions call and use the lazy evaluation of flatMap: Stream.of(supplierLamda).flatMap(Supplier::get)

In my test with a 101x101 grid of armorstands standing on the ground, this optimization brought the 145mspt down to 60mspt. In that case of non moving entities the flatMap is never evaluated (breakpoint not hit).

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/entity/stream_entity_collisions_lazily/MixinEntity.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/entity/stream_entity_collisions_lazily/MixinEntity.java
@@ -15,22 +15,6 @@ import java.util.stream.Stream;
 @Mixin(Entity.class)
 public class MixinEntity {
     /**
-     * Redirect to allow skipping getting entities from world when the stream isn't used.
-     * The resulting stream will only evaluate the getEntityCollisions lambda when it is queried.
-     *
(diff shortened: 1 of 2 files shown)


jellysquid3:
The second issue you mention looks to be caused by a performance regression introduced by Lithium non-lazily computing entity collisions [here](https://github.com/jellysquid3/lithium-fabric/blob/1.16.x/fabric/src/main/java/me/jellysquid/mods/lithium/mixin/entity/collisions/MixinEntityView.java#L26). It would be optimal to fix the issue there instead.

jellysquid3 (approved):
Code looks good to me otherwise.

2No2Name:
I agree that it is a better place

PR #44, opened 2020-05-13
Add extra collision check with block below entity to cancel momentum …

2No2Name:
Entities standing on the ground constantly try to move downwards, but usually fail to do so, because there is a block right below their feet. The block collision code iterates over the blocks around the entity, trying to collide with about 20 other blocks before actually trying to collide with the block below the entities feet.

This commit adds an early collision with the block below the feet. If all movement is cancelled the block iteration is skipped. Otherwise the normal block iteration is performed (including the block below the feet).

In my test with a 101x101 grid of armorstands standing on the ground, representing any non moving gravity affected entities, this optimization brought the 61mspt down to 54mspt.

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/entity/gravity_check_block_below/VoxelShapesMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/entity/gravity_check_block_below/VoxelShapesMixin.java
@@ -1,0 +1,44 @@
+package me.jellysquid.mods.lithium.mixin.entity.gravity_check_block_below;
+
+import net.minecraft.block.BlockState;
+import net.minecraft.block.ShapeContext;
+import net.minecraft.util.math.*;
+import net.minecraft.util.shape.VoxelShape;
+import net.minecraft.util.shape.VoxelShapes;
+import net.minecraft.world.WorldView;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import java.util.stream.Stream;
+
+@Mixin(VoxelShapes.class)
+public class VoxelShapesMixin {
+    /**
+     * Check the block below the entity first, as it is the block that is most likely going to cancel the movement from
+     * gravity.
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/me/jellysquid/mods/lithium/mixin/entity/gravity_check_block_below/MixinVoxelShapes.java
```diff
@@ -0,0 +1,43 @@
...
+public class MixinVoxelShapes {
+    /**
+     * Check the block below the entity first, as it is the block that is most likely going to cancel the movement from
+     * gravity.
+     */
+    @Inject(method = "calculatePushVelocity(Lnet/minecraft/util/math/Box;Lnet/minecraft/world/WorldView;DLnet/minecraft/block/ShapeContext;Lnet/minecraft/util/math/AxisCycleDirection;Ljava/util/stream/Stream;)D",
+            at = @At(value = "INVOKE", target = "Lnet/minecraft/util/math/AxisCycleDirection;opposite()Lnet/minecraft/util/math/AxisCycleDirection;", ordinal = 0),
+            cancellable = true, locals = LocalCapture.NO_CAPTURE)
+    private static void checkBelowFeet(Box box, WorldView world, double initial, ShapeContext context, AxisCycleDirection direction, Stream<VoxelShape> shapes, CallbackInfoReturnable<Double> cir) {
+        //                  [VanillaCopy]
```
jellysquid3: These `[VanillaCopy]` annotations do not explain what vanilla behavior they are copying and defeat the purpose of using them. Additional documentation on what exactly is going on here should be included.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/gravity_check_block_below/MixinVoxelShapes.java
```diff
@@ -0,0 +1,43 @@
...
+        if (initial >= 0 || direction.opposite().cycle(Direction.Axis.Z) != Direction.Axis.Y) {
+            return;
+        }
+        int x,y,z;
```
jellysquid3: Style nitpick: These variables should be initialized inline.

jellysquid3 (changes requested):
Code looks functional and works well in my testing, but there are a few style nitpicks I'd like to resolve before merging.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/gravity_check_block_below/MixinVoxelShapes.java
2No2Name: Okay I added a more detailed description now

2No2Name:
Force pushed to properly rebase onto the dev branch and also follow the new naming standards

PR #45, opened 2020-05-13
new: Make entities only get relevant entities for collisions by tracking custom entity class groups in chunk entity lists

2No2Name:
Similarly to the entity type predicates, the same issue happens when searching collision boxes (like boat, shulker, all entities for minecarts). Getting only the entity types that actually have collision boxes can improve performance drastically, especially when there are many entities that don't have a collision box.

Sadly the chunk entity list (TypeFilterableList) does not allow grouping entities by anything but a subclass of entity, which makes creating a group of entities with a collision box hard. This is why I replaced the internals of the TypeFilterableList and allow grouping by dynamic groups.

For the problem of other mods possibly adding entities that have hard collision boxes is solved by keeping track of which entity classes exist in the world, and analyzing the class of the entity when a new one appears. As chunkloading might happen offthread, I chose to use ConcurrentHashMaps to avoid visibility problems after loading chunks with a previously unseen type of entity.


jellysquid3:
Do you have a real-world benchmark for this change? The patches here are rather invasive and extensive... leaves a lot to test and review.

2No2Name:
I did not benchmark the changes of the typefilterable list alone, together with the collision optimization using the group of entities with collision boxes I had pretty good improvements with entities that check for entity collisions all the time. For example using about 3800 item entities in a flatworld
`/fill ~-49 1 ~-49 ~50 1 ~50 stone
/fill ~-49 1 ~-49 ~50 1 ~50 air destroy
` gave me 30mspt without this pullrequest and 22mspt with the pullrequest.

2No2Name:
These changes look quite invasive and also require more hacks to work in 1.16.1 because Minecraft code now also explicitly requires a List. I think a less invasive method should be found. Otherwise a lot of testing and profiling of several usecases has to be done to ensure that these changes don't make the performance worse.

2No2Name (commented):
Code looks good to me now. Due to the changes new performance testing is required though.

2No2Name:
Same test as before: 24 mspt to 15 mspt

src/main/java/me/jellysquid/mods/lithium/common/entity/EntityClassGroup.java
```diff
@@ -0,0 +1,119 @@
...
+    //Keep track of available class groups for updating them in case an entity class is instantiated for the first time
+    private static final List<EntityClassGroup> entityClassGroups = new ArrayList<>();
+
+    public static final EntityClassGroup COLLISION_BOX_OVERRIDE = new EntityClassGroup(
+            (Class<?> entityClass) -> {
+                boolean overwritten;
+                while(entityClass != null && entityClass != Entity.class) {
+                    try {
+                        overwritten = true;
+                        entityClass.getDeclaredMethod("method_30948");
```
Earthcomputer:
```suggestion
                        entityClass.getDeclaredMethod(FabricLoader.getInstance().getMappingResolver().mapMethodName("intermediary", "net/minecraft/class_1297", "method_30948", "()Z"));
```

src/main/java/me/jellysquid/mods/lithium/common/entity/EntityClassGroup.java
```diff
@@ -0,0 +1,119 @@
...
+                return false;
+            }
+    );
+    public static final EntityClassGroup HARD_COLLISION_BOX_OVERRIDE = new EntityClassGroup(
+        (Class<?> entityClass) -> {
+            boolean overwritten;
+            while(entityClass != null && entityClass != Entity.class) {
+                try {
+                    overwritten = true;
+                    entityClass.getDeclaredMethod("method_30949", Entity.class);
```
Earthcomputer:
```suggestion
                    entityClass.getDeclaredMethod(FabricLoader.getInstance().getMappingResolver().mapMethodName("intermediary", "net/minecraft/class_1297", "method_30949", "(Lnet/minecraft/class_1297;)Z", Entity.class);
```

2No2Name: Added that now. Thanks for the suggestion

2No2Name:
Tested the new version outside the dev environment, remapping actually works. Thanks to @Earthcomputer

PR #46, opened 2020-05-14
new: Optimize entity data tracker

maityyy:
The DataTracker `HashMap` entries map can be replaced with a faster `Int2ObjectOpenHashMap`, which optimizes the iteration process and also reduces memory consumption.


jellysquid3:
Do you have any benchmarks for this change? I'm a little bit suspicious, because get/put/contains and iterator methods in DataTracker are still being accessed through the HashMap interface, meaning those calls will box the integer keys and allocate new objects for each entry produced by the default Map iterator.

jellysquid3 (approved):
Code looks good to me, but further benchmarking would be good to see whether or not this change has a meaningful impact to justify an additional Mixin file.

maityyy:
I will check this when the new Minecraft Snapshot 20w21a comes out :)

PR #58, opened 2020-06-20
Use faster cache implementation for isShapeFullCube check

Gegy:
This change mainly affects vine generation, which calls `Block.isShapeFullCube` extensively. Because so many vines are placed in jungles, this can create a noticeable improvement in jungle generation speed.

Here's some pretty rough profiler results when generating a jungle:

Before
Before

After
After

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/common/util/collections/Object2BooleanCacheTable.java
+++ b/src/main/java/me/jellysquid/mods/lithium/common/util/collections/Object2BooleanCacheTable.java
@@ -1,0 +1,61 @@
+package me.jellysquid.mods.lithium.common.util.collections;
+
+import it.unimi.dsi.fastutil.HashCommon;
+import net.minecraft.util.math.MathHelper;
+
+import java.util.function.Predicate;
+
+/**
+ * A lossy hashtable implementation that stores a mapping between an object and a boolean.
+ *
+ * Any hash collisions will result in an overwrite: this is safe because the correct value can always be recomputed,
+ * given that the given operator is deterministic.
+ *
+ * This implementation is safe to use from multiple threads
+ */
+public final class Object2BooleanCacheTable<T> {
+    private final int capacity;
+    private final int mask;
+
+    private final Node<T>[] nodes;
+
+    private final Predicate<T> operator;
+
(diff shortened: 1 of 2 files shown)


Ueaj-Kerman (approved):
noice, have u tested different hashes/array sizes to see if u can get less collisions with vanilla blocks? (if there are any)

Gegy:
There's *a lot* of different `VoxelShape` instances created by Vanilla, unfortunately. They don't implement `.equals` either, so it's all reference equality checks. I don't think it'd be possible to avoid collisions entirely (there are a lot), but caching like this takes the edge off of repeated accesses. 
Definitely might be worth increasing the table size, though

Gegy:
I think most relevant to consider is just the ratio between cache hits vs cache misses. With a cache size of 512, on average the ratio of hits to misses is 5,000,000 to 2.

src/main/java/me/jellysquid/mods/lithium/common/util/collections/Object2BooleanCacheTable.java
```diff
@@ -0,0 +1,61 @@
...
+    public boolean get(T key) {
+        int idx = hash(key) & this.mask;
+
+        Node<T> node = this.nodes[idx];
+        if (node != null && key.equals(node.key)) {
+            return node.value;
+        }
+
+        boolean test = this.operator.test(key);
+        this.nodes[idx] = new Node<>(key, test);
```
jellysquid3, line 47: It should be possible to remove the `Node` object allocation and de-reference here by splitting the `nodes` array into two parallel arrays for the key and boolean values.

jellysquid3 (commented):
LGTM. This section of code showed up as a hotspot in a few scenarios but I never had the time to look more extensively into it. I've left a comment where I think one more additional improvement can be made to the backing cache collection if you're up for it. :smile:

jellysquid3:
Actually, ignore my comment. Doing such would make this no longer thread safe since readers could see values being updated.

jellysquid3 (approved):

Gegy:
It hurts to use allocations like this, but I don't see another way about it without synchronization. 😄
_Valhalla when_ 😔

PR #59, opened 2020-06-20
Fast caching for biome layer sampling

Gegy:
This implements a much faster cache implementation for biome layer sampling. The default one is hugely flawed in being totally wrapped in a `synchronized` block, as well as maintaining its size by removing entries one-by-one from a linked list.

This implementation alternatively works on a fixed-size "lossy" cache which evicts old entries simply by overriding them. This hugely reduces the complexity of caching, creating a substantial improvement in sampling cost. This is especially relevant for stronghold generation, which takes a huge sample area to find a spawn location. It should generally improve the performance of chunk generation too, though.

To avoid using a `synchronized` block, we also keep a ThreadLocal instance of the biome layer stack.

Furthermore, in places where a vanilla layer is used from multiple other layers, the whole stack is duplicated. This is bad for caching, because it means nothing can be shared. This PR also ensures that those layers don't get duplicated for better caching and memory usage.

Also thanks to @SuperCoder7979 & @Kroppeb :)


src/main/java/me/jellysquid/mods/lithium/mixin/world/layer_caching/MixinParentedLayer.java
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraft.world.biome.layer.util.LayerFactory;
+import net.minecraft.world.biome.layer.util.LayerSampleContext;
+import net.minecraft.world.biome.layer.util.LayerSampler;
+import org.spongepowered.asm.mixin.Mixin;
+
+/**
+ * Memoize the LayerFactory created by the ParentedLayer such that layers don't get duplicated when accessed from
+ * multiple children layers
+ *
+ * Because we can't apply mixin to default method, just override in classes that get duplicated
```
jellysquid3: I'm not too thrilled about needing to override a default method in multiple classes.

It looks like Mixins is breaking down when trying to overwrite methods in the interface due to the fact that lambdas are being used in a rather complex way. It is possible to change this class to an interface and then overwrite the defaulted method by moving this code into a static helper class elsewhere, which would in turn allow the JVM to compile this.

src/main/java/me/jellysquid/mods/lithium/common/world/layer/LayerFactoryUtil.java
```diff
@@ -0,0 +1,20 @@
...
+import net.minecraft.world.biome.layer.util.LayerSampler;
+
+public final class LayerFactoryUtil {
+    public static <A extends LayerSampler> LayerFactory<A> memoize(LayerFactory<A> factory) {
+        return new LayerFactory<A>() {
+            final ThreadLocal<A> cached = new ThreadLocal<>();
+
+            @Override
+            public A make() {
+                if (this.cached.get() == null) {
```
jellysquid3: Unlikely to be a hotspot, but the repeated calls to `ThreadLocal#get` can be avoided by using a pattern such as:
```java
T val = this.cached.get();
if (val == null) {
  this.cached.set(val = computeValue());
}
return val;
```

jellysquid3 (changes requested):
It should be possible to mixin to `ParentedLayer` with a little bit of coercing. Lithium generally tries to avoid making assumptions about the state of any class it is touching, and as such relying on these methods to not be overridden in subclasses is a bit of a no-go.

Other than nitpicks about the glue code here, the overall implementation looks like a big win over what vanilla is doing. Would you be able to provide benchmarks for this change?

jellysquid3:
It might actually be easier to remove `LayerFactoryUtil` all together and use something like [this](https://gist.github.com/jellysquid3/38772bfc5e0b6479e42ba7791da25eca) instead. I find the code much easier to follow and Mixins/JVM seems happy to deal with it.

jaskarth:
Yeah, those changes make it a lot more readable, I'll apply those. As for the performance increase with the changes, on the seed `6250325407383374817` searching for a Mushroom Fields biome through `/locatebiome minecraft:mushroom_fields` takes less than a second (~5815 blocks away) and without these changes it takes about 3 seconds on the same seed. We'll get some better profiling stats soon.

Gegy:
Ooh, yeah- that is *a lot* nicer. I was under the impression that interfaces could not be mixin'd at all, but that's super useful! 
There's also the possibility of reducing accesses to the ThreadLocal when doing 'bulk' operations (and just acquiring the layer stack once), but I feel like they could potentially be quite invasive and I'm not sure if the performance benefit would be huge.

jellysquid3 (approved):

jellysquid3:
These changes look good to me, and the rudimentary benchmarks look to show a big win here. Great work.

PR #61, opened 2020-06-25
new: faster implementation of ScaleLayer

jaskarth:
This PR replaces ScaleLayer's sample method with a faster one that avoid sampling the surrounding biomes unless it is absolutely necessary, as ScaleLayer is the most used layer in the vanilla layer stack. This optimization is not that big when it comes to world generation, but it does have a noticeable impact when the locatebiome command is used. It also optimizes `x >> 1 << 1` into `x & ~1` for a bit smarter bitwise operation. There are no worldgen differences with this enabled.

Thanks to @gegy1000 and @Kroppeb for writing this :D


Kroppeb:
It's `~1` which is `-2` and not `-1`

jaskarth:
yeah sorry, typo 😅

jellysquid3 (approved):
LGTM. Do you have any numbers for performance that you could share?

jaskarth:
Yeah. Unfortunately this isn't a super big performance improvement but when testing with a profiler (with `-XX:+UnlockDiagnosticVMOptions -XX:+DebugNonSafepoints` enabled to avoid some of the safepoint bias) and the amount of profiler samples in `ScaleLayer#sample` decreased from 48 to 38 (26% improvement). It's not a massive speed boost but it should increase the performance of the biome generator overall.

jellysquid3 (approved):

PR #63, opened 2020-06-27
Optimize the nether biome generator

jaskarth:
This PR optimizes the nether biome generator by removing the stream based code used to evaluate the biome at a position with regular collections. This optimization is most apparent when calling /locatebiome or getting every biome in a large area. In my test code to evaluate every biome in a 4096x4096 area, the optimization resulted in a 25% speed increase (from 32 seconds to 24 seconds). This PR doesn't change the generation of nether biomes at all.

Thanks again to @gegy1000 and @Kroppeb for their help ;)


jellysquid3 (approved):
LGTM. Minor style nitpicks can be solved in a merge.

PR #64, opened 2020-06-29
new: avoid sampling in the chunk generator when possible

jaskarth:
For some context, Minecraft uses two different perlin noise samplers to generate minecraft's terrain. The end result is the value of those two samplers, interpolated using another sampler. This interpolation sampler is normalized poorly, and it spends most of it's time at either 1 or 0.
image
Because the value is at the extremes, the interpolation used between the two perlin noise samplers is unnecessary and can be avoided unless the interpolation noise is within the [0, 1] range. In pseudocode, we are doing this:
```java
if (interpolation >= 1) {
    return getNoise2();
} else if (interpolation <= 0) {
    return getNoise1();
} else {
    return lerp(getNoise1(), getNoise2(), interpolation);
}
```
Some rudimentary profiling shows that this causes 33% less sampling and reduces the overhead of the chunk generator. It also preserves the worldgen shape and doesn't cause any worldgen inconsistencies from some extensive testing.

Thanks again for @gegy1000 and @Kroppeb for helping write this and find bugs in my code 😄


src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+     */
+    @Overwrite
+    private double sampleNoise(int x, int y, int z, double horizontalScale, double verticalScale, double horizontalStretch, double verticalStretch) {
+        // To generate it's terrain, Minecraft uses two different perlin noises.
+        // It interpolates these two noises to create the final sample at a position.
+        // However, the interpolation noise is not all that good and spends most of it's time at > 1 or < 0, rendering
+        // one of the noises completely unnecessary in the process.
+        // By taking advantage of that, we can reduce the sampling needed per block through the interpolation noise.
+
+        // This controls both lacunarity and persistence.
```
Gegy: Minor thing; I think rather it controls the frequency and amplitude- the lacunarity and persistence refer only to the multiplier to frequency/amplitude at each octave

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+            for (int octave = 0; octave < 16; octave++) {
+                double scaledVerticalScale = verticalScale * frequency;
+                upperNoise += sampleOctave(this.upperInterpolatedNoise.getOctave(octave), x, y, z, horizontalScale, verticalScale, scaledVerticalScale, frequency);
+                lowerNoise += sampleOctave(this.lowerInterpolatedNoise.getOctave(octave), x, y, z, horizontalScale, verticalScale, scaledVerticalScale, frequency);
+
+                frequency /= 2.0;
+            }
+
+            // Vanilla behavior, return interpolated version
+            return MathHelper.clampedLerp(lowerNoise / 512.0, upperNoise / 512, clampedInterpolation);
```
Gegy: Since we know now that the lerp value is [0; 1], we can maybe avoid some branching by using the normal `lerp`

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    private static double sampleOctave(PerlinNoiseSampler sampler, int x, int y, int z, double horizontalScale, double verticalScale, double scaledVerticalScale, double frequency) {
+        return sampler.sample(
```
Gegy: Can move the scaledVerticalScale computation into here and use it when working out the scaled y value too :)

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
jaskarth: Good catch! I seem to have mixed up those terms.

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -76,22 +73,22 @@ private double sampleNoise(int x, int y, int z, double horizontalScale, double v
...
 
-    private static double sampleOctave(PerlinNoiseSampler sampler, int x, int y, int z, double horizontalScale, double verticalScale, double scaledVerticalScale, double frequency) {
+    private static double sampleOctave(PerlinNoiseSampler sampler, int x, int y, int z, double horizontalScale, double verticalScale, double frequency) {
+        double scaledVerticalScale = verticalScale * frequency;
         return sampler.sample(
-                OctavePerlinNoiseSampler.maintainPrecision((double)x * horizontalScale * frequency),
-                OctavePerlinNoiseSampler.maintainPrecision((double)y * verticalScale * frequency),
-                OctavePerlinNoiseSampler.maintainPrecision((double)z * horizontalScale * frequency), scaledVerticalScale, (double)y * scaledVerticalScale) / frequency;
+                OctavePerlinNoiseSampler.maintainPrecision(x * horizontalScale * frequency),
+                OctavePerlinNoiseSampler.maintainPrecision(scaledVerticalScale * frequency),
```
Gegy: This should be `y * scaledVerticalScale` 😛

jaskarth: haha oops

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -0,0 +1,94 @@
...
+        } else {
+            // [VanillaCopy] SurfaceChunkGenerator#sampleNoise
+            // Sample both and interpolate, as in vanilla.
+
+            frequency = 1.0;
+            double lowerNoise = 0.0;
+            double upperNoise = 0.0;
+
+            for (int octave = 0; octave < 16; octave++) {
+                upperNoise += sampleOctave(this.upperInterpolatedNoise.getOctave(octave), x, y, z, horizontalScale, verticalScale, frequency);
```
jellysquid3: The parameters for these two `sampleOctave` calls are the same outside of their noise source. It should be possible to optimize this section quite a bit more through hoisting the `OctavePerlinNoiseSampler.maintainPrecision` calls and such. Those hoisted results can then be re-used between the two `PerlinNoiseSampler#sample` calls.

jellysquid3 (commented):
I think there's still an optimization opportunity lurking here, but I'm not sure how much of a difference it will really make. The `maintainPrecision` code looks hot and optimizing it doesn't seem trivial. But perhaps we could avoid as many calls as possible to it?

jellysquid3 (approved):

PR #65, opened 2020-06-30
new: Implement some new performance-optimizing patches

maityyy:


maityyy:
I want to apologize in advance, because I am from another country and do not know much English.

This PR introduces several new optimizations for the game and improves existing patches to squeeze extra performance.
- Replacing complex and unnecessary stream-based code with simple iterations in the complex mob AI of Piglins and Villagers, which reduces overall memory and CPU consumption on large servers
- Safely caching tameable entity owner to reduce unnecessary calls to DataTracker
- Help JVM to optimize hot functions by reducing bytecode instructions
- Eliminating overhead in vector offset functions, which speeds up chunk gen during feature/structure generation
- Fix [MC-117075](https://bugs.mojang.com/browse/MC-117075) (lag spikes during unloading large number of furnaces, chests and other block entities), which is quite often observed in large worlds and servers with a lot of players online
- Using more specifically-optimized collection for entity data tracker entries
- A very, very small improvement for ticking loaded chunks by caching the block state property, which is responsible for the ability to tick this block state, also avoiding dynamic dispatch call

build.gradle
```diff
@@ -47,9 +48,7 @@ processResources {
 	}
 }
 
-// ensure that the encoding is set to UTF-8, no matter what the system default is
-// this fixes some edge cases with special characters not displaying correctly
-// see http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
+// http://yodaconditions.net/blog/fix-for-java-file-encoding-problems-with-gradle.html
```
jellysquid3: Why was the explanation removed? If the link goes down, it will be harder to understand what is going on here. It is meant to supplement the explanation.

build.gradle
```diff
@@ -27,18 +27,19 @@ minecraft {
 }
 
 dependencies {
-	//to change the versions see the gradle.properties file
-	minecraft "com.mojang:minecraft:${project.minecraft_version}"
-	mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
-	modCompile "net.fabricmc:fabric-loader:${project.loader_version}"
+    minecraft "com.mojang:minecraft:${project.minecraft_version}"
```
jellysquid3, line 30: Indentation here has changed.

gradle.properties
```diff
@@ -1,10 +1,10 @@
-# Done to increase the memory available to gradle.
-org.gradle.jvmargs=-Xmx1G
-# Fabric Properties
-# check these on https://modmuss50.me/fabric.html
+org.gradle.jvmargs=-Xmx1G -Xms1G
```
jellysquid3: Why was this changed? The JVM should be able to grow the heap size without issue, and forcing it to always allocate the memory causes the Gradle daemon to persist with high memory usage.

src/main/java/me/jellysquid/mods/lithium/mixin/ai/task/MixinBreedTask.java
```diff
@@ -0,0 +1,54 @@
...
+
+@Mixin(BreedTask.class)
+public class MixinBreedTask {
+    @Shadow
+    @Final
+    private EntityType<? extends AnimalEntity> targetType;
+
+    /**
+     * @reason Replace stream code with traditional iteration
+     * @author Maity, JellySquid
```
jellysquid3, line 26: I do not understand why I have been added as an author here.

src/main/java/me/jellysquid/mods/lithium/mixin/ai/task/MixinBreedTask.java
```diff
@@ -0,0 +1,54 @@
...
+                final AnimalEntity animal = (AnimalEntity) mob;
+
+                if (entity.canBreedWith(animal)) {
+                    ret = animal;
+                    break;
+                }
+            }
+        }
+
+        return Optional.of(ret);
```
jellysquid3: This will throw an exception if `ret` is null, which can and probably will happen... `Optional#ofNullable` should be used instead.

src/main/java/me/jellysquid/mods/lithium/mixin/ai/goal/MixinGoalSelector.java
```diff
@@ -44,6 +44,22 @@ private void reinit(Supplier<Profiler> supplier, CallbackInfo ci) {
...
+     * @author JellySquid, Maity
+     */
+    @Overwrite
+    public void remove(Goal goal) {
+        for (PrioritizedGoal wrappedGoal : this.goals) {
+            if (wrappedGoal.getGoal() == goal) {
+                if (wrappedGoal.isRunning()) {
+                    wrappedGoal.stop();
+                }
+                this.goals.remove(wrappedGoal);
```
jellysquid3: You cannot modify a LinkedList while it is being iterated. See the javadoc [here](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedList.html) for more information.
> The iterators returned by this class's iterator and listIterator methods are fail-fast: if the list is structurally modified at any time after the iterator is created, in any way except through the Iterator's own remove or add methods, the iterator will throw a ConcurrentModificationException. Thus, in the face of concurrent modification, the iterator fails quickly and cleanly, rather than risking arbitrary, non-deterministic behavior at an undetermined time in the future. 

This loop should instead use an `Iterator` manually and call `Iterator#remove()` upon it.

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinBlockPos.java
```diff
@@ -0,0 +1,218 @@
...
+        super(x, y, z);
+    }
+
+    /**
+     * @reason Better inlining & reduce bytecode instructions
+     * @author Maity
+     */
+    @Overwrite
+    public static long asLong(int x, int y, int z) {
+        return (((long) x & (long) 67108863) << 38) | (((long) y & (long) 4095)) | (((long) z & (long) 67108863) << 12);
```
jellysquid3: What are these magic numbers?

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinChunkSectionPos.java
```diff
@@ -0,0 +1,21 @@
...
+
+@Mixin(ChunkSectionPos.class)
+public class MixinChunkSectionPos {
+    /**
+     * @reason Better inlining & reduce bytecode instructions
+     * @author Maity
+     */
+    @Overwrite
+    public static long asLong(int x, int y, int z) {
+        return (((long) x & 4194303L) << 42) | (((long) y & 1048575L)) | (((long) z & 4194303L) << 20);
```
jellysquid3: Ditto. What are these magic numbers?

src/main/java/me/jellysquid/mods/lithium/mixin/world/explosions/MixinExplosion.java
```diff
@@ -60,9 +60,7 @@
     public abstract DamageSource getDamageSource();
 
     @Shadow
-    public static float getExposure(Vec3d self, Entity entity) {
-        throw new UnsupportedOperationException();
-    }
+    public static native float getExposure(Vec3d self, Entity entity);
```
jellysquid3, line 63: Why was this changed to `native`? Is this even legal?

src/main/java/me/jellysquid/mods/lithium/mixin/world/tile_unloading/MixinServerWorld.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+import java.util.function.Supplier;
+
+@Mixin(ServerWorld.class)
+public abstract class MixinServerWorld extends World {
+    private MixinServerWorld(MutableWorldProperties props, RegistryKey<World> worldKey, RegistryKey<DimensionType> dimensionKey, DimensionType type, Supplier<Profiler> profiler, boolean isDebugWorld, boolean isClient, long seed) {
+        super(props, worldKey, dimensionKey, type, profiler, isDebugWorld, isClient, seed);
+    }
+
+    @Redirect(method = "dump", at = @At(value = "INVOKE",
```
jellysquid3: What is this code doing? I'm not sure why the backing collection is being changed here, especially given that it is a diagnostic.

src/main/java/me/jellysquid/mods/lithium/mixin/world/tile_unloading/MixinServerWorld.java
```diff
@@ -0,0 +1,35 @@
...
+            target = "Ljava/util/List;size()I"))
+    private int changeCollectionSizeLink(List<BlockEntity> blockEntities) {
+        return this.tickingBlockEntities.size();
+    }
+
+    @Redirect(method = "dumpBlockEntities", at = @At(value = "INVOKE",
```
jellysquid3: Ditto.

src/main/java/me/jellysquid/mods/lithium/mixin/world/chunk_inline_state_access/MixinPalettedContainer.java
```diff
@@ -0,0 +1,31 @@
...
+    private Palette<T> palette;
+
+    /**
+     * @reason Help JVM to optimize code by reducing instructions
+     * @author Maity
+     */
+    @Overwrite
+    public T get(int x, int y, int z) {
+        // this.get(toIndex(x, y, z))
+        T o = this.palette.getByIndex(this.data.get(y << 8 | z << 4 | x));
```
jellysquid3: Is this code really unable to be inlined by the JVM? `PalettedContainer#toIndex` is a very small static method that should be trivially inlined.

src/main/java/me/jellysquid/mods/lithium/mixin/world/chunk_inline_state_access/MixinProtoChunk.java
```diff
@@ -0,0 +1,69 @@
...
+     * @reason Help JVM to optimize code by reducing instructions
+     * @author JellySquid, Maity
+     */
+    @Overwrite
+    public BlockState getBlockState(BlockPos blockPos) {
+        final int x = blockPos.getX();
+        final int y = blockPos.getY();
+        final int z = blockPos.getZ();
+
+        if (y < 0 || y >= 256) { // [VanillaCopy] ProtoChunk#getFluidState
```
jellysquid3: Again, the JVM should be able to inline `World#isInvalidHeight(int)` here, as it's a very small static method.

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinBlockPos.java
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * This mixin patch speeds up {@link BlockPos} offset functions eliminating unnecessary overhead when
+ * calculating a new vector position. {@link BlockPos} is often used in many parts of the game code,
+ * but this patch is most noticeable during feature/structure generation.
+ *
+ * @author Maity
+ */
+@Mixin(BlockPos.class)
+public abstract class MixinBlockPos extends Vec3i {
```
jellysquid3, line 19: I have seen these optimizations before in Minecraft 1.15, and earlier versions of Lithium actually [patched it.](https://github.com/jellysquid3/lithium-fabric/blob/1.15.x/fabric/src/main/java/me/jellysquid/mods/lithium/asm/patches/DevirtualizeBlockPosTransformer.java)

Given that the de-virtualization issue is no longer a problem in Minecraft 1.16, are these patches really necessary? Could you provide microbenchmarks so we can track it?

src/main/java/me/jellysquid/mods/lithium/mixin/entity/owner_cache/MixinTameableEntity.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @author Maity
+ */
+@Mixin(TameableEntity.class)
+public abstract class MixinTameableEntity extends AnimalEntity {
+    @Shadow
+    @Final
+    private static TrackedData<Optional<UUID>> OWNER_UUID;
+
+    private UUID ownerUuid = null;
```
jellysquid3, line 28: It is possible for other mods to modify the `DataTracker` from classes which extend `TameableEntity`, and likely the case that they might try to. This code would fail if the data tracker was modified outside one of these methods and cause a bug. I'm not sure we should rely on other mods to not touch it.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/data_tracker/use_opt_collections/MixinDataTracker.java
```diff
@@ -0,0 +1,33 @@
...
+@Mixin(DataTracker.class)
+public abstract class MixinDataTracker {
+    @Mutable
+    @Shadow
+    @Final
+    private Map<Integer, DataTracker.Entry<?>> entries;
+
+    /**
+     * Uses a more specifically-optimized collection type for entity data tracker entries that provides
+     * lower memory consumption and faster iterator implementation.
```
jellysquid3: My understanding is that the existing code in `DataTracker` which iterates over the `entries` hashmap will continue to use the old `Map#iterator()` interface, which will in turn require an additional allocation per each iterated entry because of the integer boxing. Is this really speeding anything up?

jellysquid3 (changes requested):
I have multiple concerns with these patches, and I'd like to see benchmarks for many of the optimizations. It also looks like a few bugs are lurking here, and there are a few cases where I'm not sure how the game isn't immediately crashing.

jellysquid3:
I am very unlikely to accept this PR in its current state due to the size and number of patches it adds. Each PR should be kept to small features or optimizations, as large code reviews are error-prone and excessively time consuming. 

There's also no benchmarks provided (or suggested) for each one of these individual changes, making it difficult to verify the claims.

src/main/java/me/jellysquid/mods/lithium/common/world/NullBlockEntityList.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        return false;
+    }
+
+    @Override
+    public Iterator<E> iterator() {
+        throw new UnsupportedOperationException("Lithium does not support this");
```
Ueaj-Kerman: If this was taken from turtlemod, I made a PR to it before you archived it fixing an error this causes when shutting down the server, make a singleton empty iterator or iirc java / guava provides one

src/main/java/me/jellysquid/mods/lithium/mixin/ai/task/MixinFollowMobTask.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.function.Predicate;
+
+@Mixin(FollowMobTask.class)
+public class MixinFollowMobTask {
+    @Shadow
+    @Final
+    private Predicate<LivingEntity> predicate;
+
+    /**
+     * @reason Replace stream code with traditional iteration
```
Ueaj-Kerman, line 24: perhaps a forindex loop will be even faster 😉

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinChunkSectionPos.java
Ueaj-Kerman: I think they were taken from the ChunkSectionPos class, but not sure if it really improves performance

src/main/java/me/jellysquid/mods/lithium/mixin/world/explosions/MixinExplosion.java
Ueaj-Kerman: yeah it is, I do that too sometimes because making an empty method stub every time gets a bit annoying, since mcdev doesn't autogenerate method stubs for static methods iirc

src/main/java/me/jellysquid/mods/lithium/mixin/entity/data_tracker/use_opt_collections/MixinDataTracker.java
maityyy: Oh yes ... you're right.

src/main/java/me/jellysquid/mods/lithium/mixin/ai/task/MixinFollowMobTask.java
maityyy: An iterator is the fastest way to iterate `List`.

build.gradle
maityyy: It is an accident. Fixed.

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinBlockPos.java
maityyy: According to my tests, this speeds up the offset functions by ~ 10%. We simply avoid unnecessary calculations of the new coordinates of the object.

maityyy:
I deleted old unnecessary code and fixed some bugs according to the comments.

maityyy:
This set of patches reduces memory and CPU consumption by replacing unnecessary stream-based code with simpler iterations, speeds up chunk generation (removing unnecessary BlockPos coordinate offset calculations), and also optimizes the unloading a large number of block entities. On weak computers, unloading is noticeably faster (~10 fps vs ~20 fps with 80 chests), since we fix the O(n) problem of the `contains` function, which is used in `removeAll` to unload block entities.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/data_tracker/use_opt_collections/MixinDataTracker.java
maityyy: The only thing I can offer is to do such a dumb thing :joy:
```java
@Redirect(method = "someMethod", at = @At(value = "INVOKE", target = "Ljava/util/Map;someCall()"))
private Map redirect(Map oldMap) {
    return ((Int2ObjectOpenHashMap) map);
}
```
So, I removed this patch.

src/main/java/me/jellysquid/mods/lithium/mixin/world/chunk_inline_state_access/MixinPalettedContainer.java
maityyy: This copy-paste from my past code is deleted.

src/main/java/me/jellysquid/mods/lithium/mixin/ai/goal/MixinGoalSelector.java
maityyy: Oh. I hurried and made a stupid mistake. Fixed.

src/main/java/me/jellysquid/mods/lithium/mixin/math/fast_util/MixinChunkSectionPos.java
maityyy: it's my magic old code. =) Removed.

src/main/java/me/jellysquid/mods/lithium/mixin/world/tile_unloading/MixinServerWorld.java
maityyy: This patch "deletes" this collection and therefore it is necessary to replace its use. However, some mods can still use it ... I think it’s worth splitting this patch into two patches and it might be worthwhile to disable the deletion of that list (World#blockEntities) by default.

maityyy:
I'll open it with a more refined code.

PR #67, opened 2020-07-02
new: cache noise used in the end to reduce computation time

jaskarth:
This PR speeds up end noise generation by a whopping 200% on my machine by caching the noise used to calculate biomes and island density. Vanilla samples a 25x25 region of noise for these calculations, which add up very quickly as it doesn't store any of the values for later reuse. By adopting the same type of cache used in the fast layer sampler, we can use many of the previous values from old calculations and avoid recalculating the noise. I've attached some rough profiler measurements while generating terrain the end.
Before the optimization:
image
Afterwards:
image
By caching the noise, we are spending 4x less time on the noise calculation. Optimizing this more isn't possible due to the modulo operator within the code.
This change is fully compatible with vanilla generation.

Thanks to @gegy1000 for writing most of the code here 😄


src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_island_noise/MixinSurfaceChunkGenerator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import net.minecraft.util.math.noise.SimplexNoiseSampler;
+import net.minecraft.world.biome.source.BiomeSource;
+import net.minecraft.world.gen.chunk.ChunkGeneratorType;
+import net.minecraft.world.gen.chunk.SurfaceChunkGenerator;
+
+@Mixin(SurfaceChunkGenerator.class)
+public class MixinSurfaceChunkGenerator {
```
jellysquid3, line 17: This PR claims that it speeds up End generation, but isn't this also implementing the noise cache into the overworld chunk generator?

jellysquid3 (approved):
LGTM.

jaskarth:
That SurfaceChunkGenerator patch is what the end uses to calculate the density of the surrounding terrain. It's only used when `useIslandOverrideNoise` is enabled in the chunk generator config, which is only true for the end dimension's chunk generator and chunk generators added via datapacks.

src/main/java/me/jellysquid/mods/lithium/common/world/noise/SimplexNoiseCache.java
```diff
@@ -0,0 +1,130 @@
...
+                    float deltaX = gridOriginX - offsetX * GRID_SIZE;
+                    float deltaZ = gridOriginZ - offsetZ * GRID_SIZE;
+
+                    // Calculate the density at this grid cell
+                    float scaledDistance = MathHelper.sqrt(deltaX * deltaX + deltaZ * deltaZ) * distanceFactor;
+                    float densityHere = ISLAND_RADIUS - scaledDistance;
+
+                    // Try to return early if we're over the max
+                    if (densityHere > density) {
+                        if (density >= MAX) {
```
Kroppeb: Shouldn't this be `densityHere >= MAX`

jaskarth: Yeah, good catch! It shouldn't have caused any problems, but that shaves off another grid cycle 😅

src/main/java/me/jellysquid/mods/lithium/mixin/world/fast_island_noise/MixinSurfaceChunkGenerator.java
Gegy: The "surface" (??) chunk generator now functions both for the overworld and the end. I think it was a pretty recent change though (oh lol scoder replied already ignore me)

Titaniumtown (approved):
LGTM, when will this be merged?

2No2Name:
Will test the changes now. Going to merge if everything goes well

2No2Name:
LGTM. Rebasing onto 1.1.6.x/dev

Titaniumtown:
*1.16.x/dev

PR #71, opened 2020-07-03
 Fast caching for biome layer sampling (successor to #59)

Gegy:
See #59: this has since applied a fix which would cause race conditions relating to the layer context being reused over multiple threads causing the seed to mismatch


src/main/resources/lithium.mixins.json
```diff
@@ -76,6 +76,9 @@
         "world.chunk_ticking.MixinPlayerChunkWatchingManager",
         "world.chunk_ticking.MixinThreadedAnvilChunkStorage",
         "world.explosions.MixinExplosion",
+        "world.layer_caching.MixinBiomeLayerSampler",
```
jaskarth, line 80: Maybe it's best to leave the mixins off by default, as we discussed on discord

Gegy: Hmm- so just remove them from this file? or would it have to involve default values in the config?

jellysquid3:
I'm going to go ahead and merge this, and will take care of the config stuff after.

jellysquid3 (approved):

PR #74, opened 2020-07-06
Mod overrides for config values

### LINKED ISSUE #8: Add API
Hi there,
I have come across a situation where my mod is incompatible with one of the optimizations in Lithium, and I would like a way to detect whether this optimization is enabled, and give the user a friendly message to disable this optimization.
Could you provide an API to enable other mods to determine which Lithium optimizations are enabled?
Thanks! ~Earth

Earthcomputer:
Closes #8

This PR adds the ability for other mods to override lithium config values from their `fabric.mod.json`s, in particular to fix incompatibility issues, so the hassle isn't left to the end user.

The format for doing this is as follows:
```json
{
  "custom": {
    "lithium:options": {
      "chunk.no_locking": true,
      "mixin.entity.data_tracker.use_arrays": false
    }
  }
}
```
In the above example, this mod is force-enabling chunk no locking and force-disabling the entity tracker array optimization. If one mod tries to force-enable an option, and another mod tries to force-disable it, the mod which is disabling the option wins. Mod option overrides take precedence over user overrides. The mod ids of all mods force enabling/disabling each option are logged to the console.

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/common/config/LithiumConfig.java
+++ b/src/main/java/me/jellysquid/mods/lithium/common/config/LithiumConfig.java
@@ -4,6 +4,10 @@ import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.api.ModContainer;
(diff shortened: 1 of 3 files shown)


jellysquid3 (approved):
No complaints with this PR. I talked some with users on my Discord server recently regarding what would be the best way to approach this, and I think letting mods specify configuration entries for Lithium in their mod manifest is definitely the best option here.

PR #76, opened 2020-07-06
fix: wrong range calculation in NearbyEntityTracker constructor

### LINKED ISSUE #75: Creepers aren't afraid of cats (MC 1.16.1, Lithium 0.5.0, Fabric 0.8.9b203)
### Expected Behavior
Creepers should run away from cats if they are within 6 blocks of them.

### Actual Behavior
Creepers don't care about cats at all.

### Reproduction Steps
1. Spawn in a creeper
2. Spawn in a cat right next to it
3. Nothing happens

I did search a bit around and found that #13 mentions the same problem, but apparently it was fixed? Also, over on Discord, lucy mentions the same bug 2-3 weeks ago.

2No2Name:
#75 seems to be caused by the change in https://github.com/jellysquid3/lithium-fabric/commit/88beba0a70f08953501491abc5b69e68e834ffb6#diff-c7e7b07704f15208afde4c60c807139f

This PR fixes #75 .


jellysquid3:
Oops, made an oversight there. I always seem to fib this one line.

jellysquid3 (approved):

PR #77, opened 2020-07-06
fix: TagImplMixin breaks hotswap because its target does not exist

2No2Name:
TagImpl$1 seems to have moved to SetTag, but the Mixin wasn't changed. This PR changes the mixin so that it is properly applied to SetTag.

This was also causing hotswap to break, which works now again.


jellysquid3 (approved):
LGTM.

Issue #60, opened 2020-06-24
Lithium removes hitboxes inside some blocks. 


2No2Name:
### Expected Behavior
Lithium should not change any block hitboxes.

### Actual Behavior
Lithium has fewer hitboxes than vanilla, e.g. inside extended piston bases or daylight sensors.

### Reproduction Steps
Build this contraption (using soul sand, and setblock the piston then) 
In Vanilla the outcome is a) while in Lithium the outcome is b).

### Analysis
It is caused by Lithium replacing the cuboid method that creates VoxelShapes. In vanilla sometimes VoxelShapes with a BitSet are created. They have a collision box at every bit border (8th or 4th of a block) that is inside the cuboid. Lithium does not have those collision boxes inside the cuboid.
Disabling shapes.specialized_shapes works around the issue.

### Attachments
a) https://cdn.discordapp.com/attachments/648967810139029505/724030415480029284/2020-06-21_00.38.00.png
b) https://cdn.discordapp.com/attachments/648967810139029505/724030982176374956/2020-06-21_00.40.32.png

2No2Name:
Fixed in https://github.com/jellysquid3/lithium-fabric/commit/0abafdf071b5393cfc5cc6ce4a8296d2c1bb79f9

PR #79, opened 2020-07-12
fix: add hitboxes inside some voxelshapes of cuboids

2No2Name:
This PR adds collision boxes inside some blocks (e.g. extended piston base, daylight sensors) as they are in vanilla.
This fixes: Lithium removes hitboxes inside some blocks. #60 

The bug was caused by lithium not implementing the weird behavior from vanilla, where SimpleVoxelShapes with BitSetVoxelSets behave as if they have a hitbox at every 1/8th or 1/4th of a block inside them.

The problem is solved by adding a new VoxelShapeCuboidWithCollisionBoxesInside extends VoxelShapeSimpleCube that mimicks the behavior from vanilla. The overwritten VoxelShapes.cuboid(Box box) method now decides whether a VoxelShapeSimpleCube can be used or the more complex VoxelShapeCuboidWithCollisionBoxesInside needs to be used, by testing whether vanilla would create a shape with extra hitboxes inside. This probably costs some runtime, so benchmarks need to be done.

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/common/shapes/VoxelShapeAlignedCuboid.java
+++ b/src/main/java/me/jellysquid/mods/lithium/common/shapes/VoxelShapeAlignedCuboid.java
@@ -1,0 +1,129 @@
+package me.jellysquid.mods.lithium.common.shapes;
+
+import net.minecraft.util.math.AxisCycleDirection;
+import net.minecraft.util.math.Box;
+import net.minecraft.util.math.MathHelper;
+import net.minecraft.util.shape.VoxelSet;
+import net.minecraft.util.shape.VoxelShape;
+
+/**
+ * An efficient implementation of {@link VoxelShape} for a shape with one simple cuboid.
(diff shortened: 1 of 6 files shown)


2No2Name:
The PR also changes a line in VoxelShapeSimpleCube that lead to a only theoretical edge case being very slightly different from vanilla in the past. Entities seem to be able to be moved backwards when colliding with a block by at most 1e-7 if the block is the last block in the iteration of all blocks that can be collided. That does not happen realistically, because vanilla usually checks a lot of blocks around the moving entity (oversized blocks such as moving fence gates have to be considered).

2No2Name:
After some testing I found some small bugs in the implementation. Changes will come, converting to draft for now.

2No2Name:
Fixed the differences to vanilla now. Also added a test that compares the behavior of the custom voxel shape implementation to the vanilla one. Besides the VoxelShapeSimpleCube there are now two subclasses that implement the collision boxes inside the shapes and the shape being offset including the collision boxes.

PR #83, opened 2020-07-20
new: use shape cache in collision code

2No2Name:
This PR speeds up getting VoxelShapes from blocks by using the shapeCache of the blocks. Without any good reason vanilla wasn't using the shapeCache in all places.

Due to an inconsistency in vanilla code (LAVA not marked as dynamicBounds() but still changing depending on the ShapeContext (Strider entity)) an additional check for LAVA is required. There are no other blocks with that behavior.


Titaniumtown:
Compiled it and ran it on my pi. Seems faster

Titaniumtown (approved):
LGTM

2No2Name:
After some testing I found that this PR creates problems with position dependent shapes. E.g. bamboo. This cannot be merged.

Also after a simple test with some entities on the ground I could not see any significant difference in performance.

2No2Name:
Closing this pullrequest because it creates problems and doesn't help with performance anyways, so working around the problems isn't likely to be better than the vanilla implementation.

2No2Name:
This can be merged when https://bugs.mojang.com/browse/MC-196298 is fixed in vanilla, or we fix it ourselves

PR #84, opened 2020-07-20
new: add chunk sections counting oversized blocks and reduce number of blocks collision is checked against

2No2Name:
This PR adds oversized block counting to chunk sections and an performance improvement based on that information.
If a chunk section has no oversized blocks (larger than full cube / no shape cache) it won't be necessary to check all the blocks around an entity while moving, as all blocks are known to be at most 1 block big. Of course this is checked seperately for each chunk section around the entity.
This PR also improves the block collision sweeper to test all necessary blocks in one chunk section at a time, allowing it to not repeat the above oversized block count check and also allowing it to cache the Chunk and ChunkSection until moving to checking collisions in the next one.
For example with item entites there was a visible improvement: https://cdn.discordapp.com/attachments/648967810139029505/713070457196249088/Screenshot_at_2020-05-21_18-47-00.png

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/common/entity/LithiumEntityCollisions.java
+++ b/src/main/java/me/jellysquid/mods/lithium/common/entity/LithiumEntityCollisions.java
@@ -1,6 +1,6 @@
 package me.jellysquid.mods.lithium.common.entity;
 
-import me.jellysquid.mods.lithium.common.entity.movement.BlockCollisionSweeper;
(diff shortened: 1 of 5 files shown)


2No2Name:
Skipping the iteration on some blocks may change entity movement by 1e-7 in rare edge cases. Also see #79 

Edit: this does not seem to happen. Vanilla also does not return VoxelShapes that are not colliding with the box that we iterate on.

2No2Name:
Adressed the 1e-7 edge case in the latest commit

PR #88, opened 2020-07-25
Speed up locating structures greatly

TelepathicGrunt:
This PR adds a single if statement into StructrueFeature's locateStructure method so that instead of generating empty chunks to check for if the structure can spawn, it checks the biome first which is far faster. 

You can see the speed boost in the videos attached to the issue report here:
https://github.com/jellysquid3/lithium-fabric/issues/85

The main line of code responsible for the significant speed boost
`if(worldView.getBiomeAccess().getBiomeForNoiseGen(chunkPos.x << 2, 60, chunkPos.z << 2).hasStructureFeature(thisStructure)) {`
                
I have tested in both vanilla and with the patch and /locate finds the exact same structure in the same spot so vanilla parity is preserved. I have agreed to the CLA and the relicensing of this piece of code. I hope this helps!

Things that will be greatly benefit from the locating structure speed boost:
-/locate command
-Treasure maps (especially those that skip existing chunks)
-Any modded item, command, mobs, etc that uses locateStructure in backend to find a structure.

The biggest noticeable change in the speed of finding a structure is doing any of the above for a structure that has a spacing of 1, separation of 0, and is added to a very rare far away biome. Doing /locate on that can hang the server for minutes on end while with this patch, will take only a few seconds instead. :)

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/gen/features/StructureFeatureMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/gen/features/StructureFeatureMixin.java
@@ -1,74 +1,70 @@
 package me.jellysquid.mods.lithium.mixin.gen.features;
 
 import net.minecraft.structure.StructureStart;
-import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.ChunkPos;
 import net.minecraft.util.math.ChunkSectionPos;
+import net.minecraft.world.StructureHolder;
 import net.minecraft.world.WorldView;
 import net.minecraft.world.chunk.Chunk;
 import net.minecraft.world.chunk.ChunkStatus;
-import net.minecraft.world.gen.ChunkRandom;
 import net.minecraft.world.gen.StructureAccessor;
-import net.minecraft.world.gen.chunk.StructureConfig;
-import net.minecraft.world.gen.feature.FeatureConfig;
 import net.minecraft.world.gen.feature.StructureFeature;
+import org.objectweb.asm.Opcodes;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Overwrite;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+import org.spongepowered.asm.mixin.injection.Slice;
 
 @Mixin(StructureFeature.class)
 public class StructureFeatureMixin {
-
     /**
-     * @reason Why generate an empty chunk to check for a structure if
+     * Why generate an empty chunk to check for a structure if
      * the chunk's biome cannot generate the structure anyway?
      * Checking the biome first = SPEED!
+     *
      * @author TelepathicGrunt
      */
-    @Overwrite
-    public BlockPos locateStructure(WorldView worldView, StructureAccessor structureAccessor, BlockPos blockPos, int i, boolean skipExistingChunks, long l, StructureConfig structureConfig) {
-        int j = structureConfig.getSpacing();
-        int k = blockPos.getX() >> 4;
-        int m = blockPos.getZ() >> 4;
-        int n = 0;
-        StructureFeature thisStructure = ((StructureFeature) (Object) this);
 
-        for(ChunkRandom chunkRandom = new ChunkRandom(); n <= i; ++n) {
-            for(int o = -n; o <= n; ++o) {
-                boolean bl = o == -n || o == n;
+    /**
+     * @reason Return null chunk if biome doesn't match structure
+     * @author MrGrim
+     */
+    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldView;getChunk(IILnet/minecraft/world/chunk/ChunkStatus;)Lnet/minecraft/world/chunk/Chunk;", ordinal = 0),
+            slice = @Slice(from = @At(value = "FIELD", opcode = Opcodes.GETSTATIC, target = "Lnet/minecraft/world/chunk/ChunkStatus;STRUCTURE_STARTS:Lnet/minecraft/world/chunk/ChunkStatus;", ordinal = 0),
+                    to = @At(value = "INVOKE", target = "Lnet/minecraft/world/chunk/Chunk;getPos()Lnet/minecraft/util/math/ChunkPos;", ordinal = 0)))
+    private Chunk biomeConditionalGetChunk(WorldView worldView, int x, int z, ChunkStatus status)
+    {
+        if (worldView.getBiomeAccess().getBiomeForNoiseGen(x << 2, 60, z << 2).hasStructureFeature((StructureFeature) (Object) this))
+            return worldView.getChunk(x, z, status);
+        else
+            return null;
+    }
 
-                for(int p = -n; p <= n; ++p) {
-                    boolean bl2 = p == -n || p == n;
-                    if (bl || bl2) {
-                        int q = k + j * o;
-                        int r = m + j * p;
-                        ChunkPos chunkPos = thisStructure.method_27218(structureConfig, l, chunkRandom, q, r);
-                        if(worldView.getBiomeForNoiseGen(chunkPos.x << 2, 60, chunkPos.z << 2).hasStructureFeature(thisStructure)) {
-                            Chunk chunk = worldView.getChunk(chunkPos.x, chunkPos.z, ChunkStatus.STRUCTURE_STARTS);
-                            StructureStart<?> structureStart = structureAccessor.getStructureStart(ChunkSectionPos.from(chunk.getPos(), 0), thisStructure, chunk);
-                            if (structureStart != null && structureStart.hasChildren()) {
-                                if (skipExistingChunks && structureStart.isInExistingChunk()) {
-                                    structureStart.incrementReferences();
-                                    return structureStart.getPos();
-                                }
+    /**
+     * @reason Can't avoid the call to Chunk.getPos(), and now the
+     * chunk might be null. Send a new (0,0) ChunkPos if so. It
+     * won't be used anyway.
+     * @author MrGrim
+     */
+    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/chunk/Chunk;getPos()Lnet/minecraft/util/math/ChunkPos;", ordinal = 0),
+            slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldView;getChunk(IILnet/minecraft/world/chunk/ChunkStatus;)Lnet/minecraft/world/chunk/Chunk;", ordinal = 0),
+                    to = @At(value = "INVOKE", target = "Lnet/minecraft/util/math/ChunkSectionPos;from(Lnet/minecraft/util/math/ChunkPos;I)Lnet/minecraft/util/math/ChunkSectionPos;", ordinal = 0)))
+    private ChunkPos checkForNull(Chunk chunk)
+    {
+        return chunk == null ? new ChunkPos(0, 0) : chunk.getPos();
+    }
 
-                                if (!skipExistingChunks) {
-                                    return structureStart.getPos();
-                                }
-                            }
-                        }
-
-                        if (n == 0) {
-                            break;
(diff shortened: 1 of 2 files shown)


src/main/java/me/jellysquid/mods/lithium/mixin/gen/features/StructureFeatureMixin.java
```diff
@@ -0,0 +1,74 @@
...
+@Mixin(StructureFeature.class)
+public class StructureFeatureMixin {
+
+    /**
+     * @reason Why generate an empty chunk to check for a structure if
+     * the chunk's biome cannot generate the structure anyway?
+     * Checking the biome first = SPEED!
+     * @author TelepathicGrunt
+     */
+    @Overwrite
```
jaskarth: If it's a simple one line change, wouldn't it be better for this to be an injector or a redirector?

jaskarth (changes requested):

src/main/java/me/jellysquid/mods/lithium/mixin/gen/features/StructureFeatureMixin.java
TelepathicGrunt: Replying here for other see as well and offer ideas.

The issue is we are in 2 nested for loops and need to skip this entire chunk of code if the biome is invalid. i couldn't figure out how to make an inject mixin to do this but maybe someone has an idea
```
Chunk chunk = worldView.getChunk(chunkPos.x, chunkPos.z, ChunkStatus.STRUCTURE_STARTS);
                            StructureStart<?> structureStart = structureAccessor.getStructureStart(ChunkSectionPos.from(chunk.getPos(), 0), thisStructure, chunk);
                            if (structureStart != null && structureStart.hasChildren()) {
                                if (skipExistingChunks && structureStart.isInExistingChunk()) {
                                    structureStart.incrementReferences();
                                    return structureStart.getPos();
                                }

                                if (!skipExistingChunks) {
                                    return structureStart.getPos();
                                }
                            }
```

TelepathicGrunt: Some ideas kicked around. hmm 
There might not be a better clean idea other than a overwrite. rip

image

mrgrim:
This is a little messy, but more targeted. It found a mansion over 35k blocks away in no time. It could use more testing to verify it finds the same thing vanilla does.

```java
public class StructureFeatureMixin {
    /**
     * Why generate an empty chunk to check for a structure if
     * the chunk's biome cannot generate the structure anyway?
     * Checking the biome first = SPEED!
     * 
     * @author TelepathicGrunt
     */

    /**
     * @reason Return null chunk if biome doesn't match structure
     * @author MrGrim
     */
    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldView;getChunk(IILnet/minecraft/world/chunk/ChunkStatus;)Lnet/minecraft/world/chunk/Chunk;", ordinal = 0),
              slice = @Slice(from = @At(value = "FIELD", opcode = Opcodes.GETSTATIC, target = "Lnet/minecraft/world/chunk/ChunkStatus;STRUCTURE_STARTS:Lnet/minecraft/world/chunk/ChunkStatus;", ordinal = 0),
                             to = @At(value = "INVOKE", target = "Lnet/minecraft/world/chunk/Chunk;getPos()Lnet/minecraft/util/math/ChunkPos;", ordinal = 0)))
    private Chunk biomeConditionalGetChunk(WorldView worldView, int x, int z, ChunkStatus status)
    {
        if (worldView.getBiomeAccess().getBiomeForNoiseGen(x << 2, 60, z << 2).hasStructureFeature((StructureFeature) (Object) this))
            return worldView.getChunk(x, z, status);
        else
            return null;
    }

    /**
     * @reason Can't avoid the call to Chunk.getPos(), and now the
     * chunk might be null. Send a new (0,0) ChunkPos if so. It
     * won't be used anyway.
     * @author MrGrim
     */
    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/chunk/Chunk;getPos()Lnet/minecraft/util/math/ChunkPos;", ordinal = 0),
              slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldView;getChunk(IILnet/minecraft/world/chunk/ChunkStatus;)Lnet/minecraft/world/chunk/Chunk;", ordinal = 0),
                             to = @At(value = "INVOKE", target = "Lnet/minecraft/util/math/ChunkSectionPos;from(Lnet/minecraft/util/math/ChunkPos;I)Lnet/minecraft/util/math/ChunkSectionPos;", ordinal = 0)))
    private ChunkPos checkForNull(Chunk chunk)
    {
        return chunk == null ? new ChunkPos(0, 0) : chunk.getPos();
    }

    /**
     * @reason Return null here if the chunk is null. This will
     * bypass the following if statement allowing the search
     * to continue.
     * @author MrGrim
     */
    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/gen/StructureAccessor;getStructureStart(Lnet/minecraft/util/math/ChunkSectionPos;Lnet/minecraft/world/gen/feature/StructureFeature;Lnet/minecraft/world/StructureHolder;)Lnet/minecraft/structure/StructureStart;", ordinal = 0),
              slice = @Slice(from = @At(value = "INVOKE", target = "Lnet/minecraft/util/math/ChunkSectionPos;from(Lnet/minecraft/util/math/ChunkPos;I)Lnet/minecraft/util/math/ChunkSectionPos;", ordinal = 0),
                             to = @At(value = "INVOKE", target = "Lnet/minecraft/structure/StructureStart;hasChildren()Z", ordinal = 0)))
    private StructureStart<?> checkChunkBeforeGetStructureStart(StructureAccessor structureAccessor, ChunkSectionPos sectionPos, StructureFeature<?> thisStructure, StructureHolder chunk)
    {
        return chunk == null ? null : structureAccessor.getStructureStart(sectionPos, thisStructure, chunk);
    }
}
```

TelepathicGrunt:
While it seems like every time I test the sped-up version of locateStructure with both my fix and mrgrim's, it always match vanilla's locate positions. Like every single time! Except for one Jungle Temple in the seed `8025490285463120952` at `/tp @s -18624 ~ -15296`

It's really odd. When the mixin is off, locate can find the jungle temple when you're right on it. But with the mixin on, it returns a different temple 300 blocks away. 

If we can figure out why this happens and fix it, we would match vanilla parity 100%! 

Edit: to be clear, this jungle temple eludes the original mixin and mrgrim's. I suspect the biome thing might be barely off but idk why it would be. The jungle temple's chunk is partially a river biome. hmm

Zirvac:
Like TelepathicGrunt, I have a case where this new method doesn't find the same structure than in vanilla.

On the seed `3721825697576006751`, the nearest desert temple is on a forest/desert frontier and the coordinates returned by the `locat` command are in the forest.

2No2Name:
I converted this to a draft until the issue above is fixed, as this cannot be merged with the difference from vanilla.

ontrigger:
I think I've found the issue here. In the case of grunt's example with the jungle temple, the biome found by getBiomeForNoiseGen was minecraft.river but the actual biome at the structure start position was minecraft.jungle_hills.

I fixed this by offsetting the pos by 2, just like in ChunkGenerator$setStructureStarts:

`worldView.getBiomeForNoiseGen((chunkPos.x << 2) + 2, 0, (chunkPos.z << 2) + 2)`

Tested it on both the jungle and desert temple seed and it seemed to locate them correctly

Kroppeb:
> I think I've found the issue here. In the case of grunt's example with the jungle temple, the biome found by getBiomeForNoiseGen was minecraft.river but the actual biome at the structure start position was minecraft.jungle_hills.
> 
> I fixed this by offsetting the pos by 2, just like in ChunkGenerator$setStructureStarts:
> 
> `worldView.getBiomeForNoiseGen((chunkPos.x << 2) + 2, 0, (chunkPos.z << 2) + 2)`
> 
> Tested it on both the jungle and desert temple seed and it seemed to locate them correctly

Does mc always add `2` too or will it use all biomes in the chunk?

ontrigger:
It only adds 2 for structure starts and gets the biome at that pos.

TelepathicGrunt:
@ontrigger Holy smokes! If this right, I own you one! I have tried `(chunkPos.x << 2 + 2, 0, (chunkPos.z << 2 + 2)` when I tried random offsets to see if that was the issue but I think I forgotten about order of operations or something lol

You're a lifesaver as this helps my mod too as I didn't do that offset

TelepathicGrunt:
Yep, the fix works! I tested with the two structures and many more and they all match up now

2No2Name:
Can someone confirm this also works with nether and end structures, or does this only affect the overworld?

Zirvac:
Nether and End seems OK.

In `-3301254986155535838` the nether fossile in `3040 ~ 2592` was overlooked by the command in the precedent commit, it's not the case with the last.

TelepathicGrunt:
Yeah I can confirm that the nether and end seem good too. I checked a bunch of structures and they all match up.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/features/StructureFeatureMixin.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * @reason Return null chunk if biome doesn't match structure
+     * @author MrGrim
+     */
+    @Redirect(method = "locateStructure", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/WorldView;getChunk(IILnet/minecraft/world/chunk/ChunkStatus;)Lnet/minecraft/world/chunk/Chunk;", ordinal = 0),
+            slice = @Slice(from = @At(value = "FIELD", opcode = Opcodes.GETSTATIC, target = "Lnet/minecraft/world/chunk/ChunkStatus;STRUCTURE_STARTS:Lnet/minecraft/world/chunk/ChunkStatus;", ordinal = 0),
+                    to = @At(value = "INVOKE", target = "Lnet/minecraft/world/chunk/Chunk;getPos()Lnet/minecraft/util/math/ChunkPos;", ordinal = 0)))
+    private Chunk biomeConditionalGetChunk(WorldView worldView, int x, int z, ChunkStatus status)
+    {
+        if (worldView.getBiomeForNoiseGen((x << 2) + 2, 0, (z << 2) + 2).hasStructureFeature((StructureFeature) (Object) this))
```
2No2Name, line 37: After rebasing onto the 1.16.2 dev branch, the hasStructureFeature method does not seem to exist. How can this be fixed?

TelepathicGrunt: It’s now 
worldView.getBiomeForNoiseGen((chunkPos.x << 2) + 2, 60, (chunkPos.z << 2) + 2).getGenerationSettings().hasStructureFeature((StructureFeature)(Object)this)

2No2Name: Okay thank you, but where does that 60 come from? ChunkGenerator.setStructureStarts doesn't have it

TelepathicGrunt: The y value doesn’t matter. I just chose 60 as default but if you follow getBiomeForNoiseGen, it’ll just use the x and z values only if I recall. Keep it 0 for now

jaskarth: Actually, in the nether the generator uses x, z, *and* y. I think 0 would be a better choice here, to keep a good baseline.

2No2Name: In the code I found calls with y=0, y=2, y = getSeaLevel(). Structure code seems to use 0

2No2Name:
I am rebasing the code onto 1.16.x/dev and testing it again now.

2No2Name:
Merged https://github.com/jellysquid3/lithium-fabric/commit/6608447a6e140340ed01437fc18b22a2994c085d

PR #90, opened 2020-07-27
new: cache netherrack blob generation

jaskarth:
In the nether, the Basalt Deltas biome generates blobs of Basalt and Blackstone to replace netherrack. These blobs iterate outwards from a point and replace the surrounding netherrack in a sphere. However, the sphere is calculated every time the feature generates, which is *100 times per chunk* leading to slowdowns. The fix here is to cache the delta positions of the blob and then reuse it, meaning once a blob of a certain size has been generated, the rest of the blobs can reuse those positions in the future, leading to an immense **83x speedup.**

These times were taken over a period of 3 minutes when the server was constantly generating chunks at the max spectator mode speeds with a mod that set Basalt Deltas to the only biome in the nether.
Before the optimization, 10% of the *total cpu time* was spent on the NetherrackReplaceBlobsFeature:
image
After the optimization, that number was reduced to 0.12%:
image

This change shouldn't change worldgen more than the existing variation between seeds that exists because of the chunk generation order.

I'd also like to thank @gegy1000 for her help with this optimization :D

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
@@ -1,0 +1,120 @@
+package me.jellysquid.mods.lithium.mixin.gen.fast_netherrack_replacement_blobs;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Random;
+
+import com.google.common.collect.ImmutableList;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Overwrite;
+import org.spongepowered.asm.mixin.Shadow;
+
+import net.minecraft.block.Block;
+import net.minecraft.block.BlockState;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
+import net.minecraft.util.math.Vec3i;
+import net.minecraft.world.ServerWorldAccess;
+import net.minecraft.world.WorldAccess;
+import net.minecraft.world.gen.StructureAccessor;
+import net.minecraft.world.gen.chunk.ChunkGenerator;
+import net.minecraft.world.gen.feature.NetherrackReplaceBlobsFeature;
+import net.minecraft.world.gen.feature.NetherrackReplaceBlobsFeatureConfig;
+
+@Mixin(NetherrackReplaceBlobsFeature.class)
+public abstract class NetherrackReplaceBlobsFeatureMixin {
+	private static final Long2ObjectOpenHashMap<List<BlockPos>> DELTA_CACHE = new Long2ObjectOpenHashMap<>();
+
+	static {
+		DELTA_CACHE.defaultReturnValue(null);
+	}
+
+	// Finds start position
+	@Shadow
+	protected static BlockPos method_27107(WorldAccess world, BlockPos.Mutable mutable, Block block) {
+		return null;
+	}
+
+	// Generates the size of this blob
+	@Shadow
+	protected static Vec3i method_27108(Random random, NetherrackReplaceBlobsFeatureConfig config) {
+		return null;
+	}
+
+	/**
+	 * @reason Faster implementation which utilizes a cache of delta positions instead of iterating outwards many times.
+     * This saves many CPU cycles as this feature is called 100 times per chunk in the Basalt Deltas biome.
+	 * @author SuperCoder79
+	 */
+	@Overwrite
+	public boolean generate(ServerWorldAccess world, StructureAccessor structureAccessor, ChunkGenerator chunkGenerator, Random random, BlockPos blockPos, NetherrackReplaceBlobsFeatureConfig config) {
+		// [VanillaCopy] NetherrackReplaceBlobsFeature#generate
+
+	    Block targetBlock = config.target.getBlock();
+
+		// Get start position
+		BlockPos startPos = method_27107(world, blockPos.mutableCopy().method_27158(Direction.Axis.Y, 1, world.getHeight() - 1), targetBlock);
+
+		// Exit early if we couldn't create the start position.
+		if (startPos == null) {
+			return false;
+		} else {
+			boolean didPlace = false;
+
+			// Get the delta positions from size vector
+			List<BlockPos> deltas = getDeltas(method_27108(random, config));
+			BlockPos.Mutable mutable = new BlockPos.Mutable();
+
+			// Iterate through the delta positions and attempt to place each one.
+			for (BlockPos delta : deltas) {
+				// Add the delta to the start position
+				mutable.set(startPos, delta.getX(), delta.getY(), delta.getZ());
+
+				// If the block here is the target, replace.
+				BlockState hereState = world.getBlockState(mutable);
+				if (hereState.isOf(targetBlock)) {
+					world.setBlockState(mutable, config.state, 3);
+					didPlace = true;
+				}
+			}
+
+			return didPlace;
+		}
+	}
+
+	private static List<BlockPos> getDeltas(Vec3i size) {
+	    // Attempt to retrieve values from the cache
+		long packed = BlockPos.asLong(size.getX(), size.getY(), size.getZ());
+		List<BlockPos> deltas = DELTA_CACHE.get(packed);
+
+		// Cache hit, return value
+		if (deltas != null) {
+			return deltas;
+		}
+
+		// Cache miss, compute and store
+
+        // Get the largest of the sides
+		int maxExtent = Math.max(size.getX(), Math.max(size.getY(), size.getZ()));
+
+		// Iterate outwards and add the deltas that are within the max extent
+		deltas = new ArrayList<>();
+		for (BlockPos pos : BlockPos.iterateOutwards(BlockPos.ORIGIN, size.getX(), size.getY(), size.getZ())) {
+			BlockPos delta = pos.toImmutable();
+
+			// Only add the positions whose distance from the origin are less than the maxExtent, creating a sphere.
+            // In vanilla, this is run every time the feature generates but we can cache it for a large benefit.
+			if (!(delta.getManhattanDistance(BlockPos.ORIGIN) > maxExtent)) {
+				deltas.add(delta);
+			}
+		}
+
+		// Store values and return
+		DELTA_CACHE.put(packed, deltas);
+
+		return deltas;
+
+	}
+}
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
@@ -1,12 +1,11 @@
 package me.jellysquid.mods.lithium.mixin.gen.fast_netherrack_replacement_blobs;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Random;
 
-import com.google.common.collect.ImmutableList;
-import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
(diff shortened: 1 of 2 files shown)


maityyy:
I think a little work needs to be done on the style of the code.

jaskarth:
More details please? I'd like to make this code as maintainable as I can 😅

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,120 @@
...
+
+			// Only add the positions whose distance from the origin are less than the maxExtent, creating a sphere.
+            // In vanilla, this is run every time the feature generates but we can cache it for a large benefit.
+			if (!(delta.getManhattanDistance(BlockPos.ORIGIN) > maxExtent)) {
+				deltas.add(delta);
+			}
+		}
+
+		// Store values and return
+		DELTA_CACHE.put(packed, deltas);
```
jellysquid3: Entries are never being removed from this cache, leading to a memory leak.

jellysquid3 (changes requested):
This PR currently has a memory leak, and it's not clear that the static field caching these entries is truly safe from other threads.

If possible, we should try to clear all old entries as soon as we know the feature generator has moved onto another chunk, but I'm not sure if this is possible to signal to the generator. Failing that, using a linked map and removing entries upon reaching a certain threshold size would work, but it'll contribute a non-trivial amount to the heap size.


I also discussed on Discord that using a `LongArrayList` over a `List<BlockPos>` would greatly improve memory compactness and eliminate almost all cache misses, helping speed this along further.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+public abstract class NetherrackReplaceBlobsFeatureMixin {
+	private static final Long2ReferenceLinkedOpenHashMap<LongArrayList> DELTA_CACHE = new Long2ReferenceLinkedOpenHashMap<>();
+
+	static {
+		DELTA_CACHE.defaultReturnValue(null);
+	}
+
+	// Finds start position
+	@Shadow
+	protected static BlockPos method_27107(WorldAccess world, BlockPos.Mutable mutable, Block block) {
```
Kroppeb: Since I have been using `native` for `@Shadow` recently, I was really confused how this was a valid optimization until I realized it's just a stub :p

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+
+		// Cache miss, compute and store
+
+        // Get the largest of the sides
+		int maxExtent = Math.max(size.getX(), Math.max(size.getY(), size.getZ()));
+
+		// Iterate outwards and add the deltas that are within the max extent
+		deltas = new LongArrayList();
+		for (BlockPos pos : BlockPos.iterateOutwards(BlockPos.ORIGIN, size.getX(), size.getY(), size.getZ())) {
+			BlockPos delta = pos.toImmutable();
```
Kroppeb: You could delay the creation of the immutable to after the distance check for that little extra performance.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			return didPlace;
+		}
+	}
+
+	private static LongArrayList getDeltas(Vec3i size) {
+	    // Attempt to retrieve values from the cache
+		long packed = BlockPos.asLong(size.getX(), size.getY(), size.getZ());
+        LongArrayList deltas = DELTA_CACHE.get(packed);
```
Kroppeb: Adding a `ReadWriteLock` should be fairly simple here.
Also the indentation is wrong here.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			return false;
+		} else {
+			boolean didPlace = false;
+
+			// Get the delta positions from size vector
+            LongArrayList deltas = getDeltas(method_27108(random, config));
+			BlockPos.Mutable mutable = new BlockPos.Mutable();
+
+			// Iterate through the delta positions and attempt to place each one.
+			for (long delta : deltas) {
```
Gegy: afaik this will use the standard Java `Iterator<Long>` and box/unbox as it goes. Might be better to use `LongIterator` directly with `.nextLong()`

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
Gegy: Since it's now being converted into a long, it shouldn't need to be made into an immutable pos at all

Gegy:
Is it time for another lossy cache map implementation? 🤠

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		// Cache miss, compute and store
+
+        // Get the largest of the sides
+		int maxExtent = Math.max(size.getX(), Math.max(size.getY(), size.getZ()));
+
+		// Iterate outwards and add the deltas that are within the max extent
+		deltas = new LongArrayList();
+		for (BlockPos pos : BlockPos.iterateOutwards(BlockPos.ORIGIN, size.getX(), size.getY(), size.getZ())) {
```
gegy1000, line 103: It's interesting that MC uses `iterateOutwards` at all. It doesn't seem like the order's important, it'd be much faster to use the regular `iterate`

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			BlockPos delta = pos.toImmutable();
+
+			// Only add the positions whose distance from the origin are less than the maxExtent, creating a sphere.
+            // In vanilla, this is run every time the feature generates but we can cache it for a large benefit.
+			if (!(delta.getManhattanDistance(BlockPos.ORIGIN) > maxExtent)) {
```
Gegy: This is slightly different to vanilla behaviour- it instead `breaks` when this condition fails

Kroppeb:
> Is it time for another lossy cache map implementation?

No as the meeting code shouldn't run in vanilla. It could run when a mod (or maybe even with custom creature configs). Making it lossy might cause a conflict and a lot more calculations.

Gegy:
The cache is already technically lossy by removing old elements when passing the certain capacity.

Kroppeb:
> The cache is already technically lossy by removing old elements when passing the certain capacity.

But that's something that shouldn't happen.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+
+		// Store values and return
+		DELTA_CACHE.put(packed, deltas);
+
+		// Ensure that the cache doesn't overflow by removing entries after a certain threshold.
+        // In this case, the chosen threshold is 125 as vanilla only creates blobs that have a size from 3-7 on all axes.
+		if (DELTA_CACHE.size() > 125) {
+		    DELTA_CACHE.removeLast();
```
Kroppeb: This drops the item you just calculated. If this is what you want you could just check the size and only add if it's less. This would be more performant.

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_netherrack_replacement_blobs/NetherrackReplaceBlobsFeatureMixin.java
maityyy: Lock can be bad for performance.  Especially on Windows systems.

Kroppeb: Yes but this isn't thread safe atm. We also made a few lossy thread safe caches already he could use but ideally it would be able to contain all 125 vanilla keys at the same time.

maityyy: > Yes but this isn't thread safe atm. We also made a few lossy thread safe caches already he could use but ideally it would be able to contain all 125 vanilla keys at the same time.

If this patch has write lock, it will be an extremely specific optimization and no one will know exactly where it will work well and where it will not work at all. We may have to disable it initially. 😔

mrgrim:
This area of the code has significant differences in 1.16.2 that you may want to get a jump on. It looks like parts of it were put into a UniformIntDistribution class, and it also looks like generation may have changed. It seems to pull a single random number for all 3 axis instead of 3 separate ones.

2No2Name:
Is this PR still up to date, or did the vanilla code change?

Edit: as the change is probably from a vanilla bug, which has been reported and marked as "important" priority, we expect the vanilla code to change soon again. Until then, we will wait

2No2Name:
I doubt this implementation is threadsafe, as there does not seem to be a guarantee that readers of the cache can see the correct state of the LongArrayLists that were inserted by a different thread into the cache. I might be wrong about this, so please explain it, if you know it.

Edit:
The code does not seem to be run by several threads without synchronization

Gegy:
~~@2No2Name: The implementation should be thread-safe. It can handle fine if multiple threads access the same entry, as it will only be inserted into the table once it is completely populated. The result is also deterministic, so values getting overridden won't cause issues.~~

TelepathicGrunt:
Wait so is this threadsafe or not? I ask as features needs to be threadsafe or else they will crash the server. What I have found is that if a mod uses a randompatchfeature instance and the feature is in two separate dimensions, the server can load the two world's chunk at the exact same time if players are exploring there. And it turns out randompatchfeature is not threadsafe and will crash the server.

So if this PR isn't threadsafe, that means if other mods or possibly datapacks uses this same feature, there's a chance the server could crash or get unexpected behavior when it loads the chunks at exactly the same time (each worlds seems to be on it's own thread?)

More testing would have to be done of course

Kroppeb:
> Wait so is this threadsafe or not? I ask as features needs to be threadsafe or else they will crash the server. What I have found is that if a mod uses a randompatchfeature instance and the feature is in two separate dimensions, the server can load the two world's chunk at the exact same time if players are exploring there. And it turns out randompatchfeature is not threadsafe and will crash the server.
> 
> So if this PR isn't threadsafe, that means if other mods or possibly datapacks uses this same feature, there's a chance the server could crash or get unexpected behavior when it loads the chunks at exactly the same time (each worlds seems to be on it's own thread?)
> 
> More testing would have to be done of course

In practise yes. But technically the JVM doesn't guarantee that different threads are seeing the cached data correctly in the current code.

jaskarth:
I'm closing this to redo the patch for 1.16.4 and make it threadsafe.

PR #94, opened 2020-07-29
Cache slime chunks

eatmyvenom:
Okay this is kinda stable at this point but I need feedback on this so it can be better. 

Implementation:
When a slime spawns by, by default a chunk is marked as a slime chunk. It then makes sure that the chunk is really supposed to be a slime chunk by double checking the random. For 1 out of 10 chunks the value will be correct by default. But otherwise it will set the value to be false. The next time it tries to get that chunk, the value will be false and it won't do the extra processing to double check. This means that it's the same as vanilla 1 out of 10 times but after a chunk has been tried.

Don't use this test in a world that has already been generated because I have no clue how that you behave.
If you have any input at all pls giv to me


src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkStatusMixin.java
```diff
@@ -0,0 +1,42 @@
...
+import net.minecraft.structure.StructureManager;
+import net.minecraft.util.math.ChunkPos;
+import net.minecraft.world.ServerWorldAccess;
+import net.minecraft.world.chunk.Chunk;
+import net.minecraft.world.chunk.ChunkStatus;
+
+@Mixin(ChunkStatus.class)
+public class ChunkStatusMixin {
+    @Inject(at = @At("RETURN"), method = "runGenerationTask")
+    public void genSlimeChunks(ServerWorld world, ChunkGenerator chunkGenerator, StructureManager structureManager, ServerLightingProvider lightingProvider, Function<Chunk, CompletableFuture<Either<Chunk, ChunkHolder.Unloaded>>> function, List<Chunk> chunks, CallbackInfoReturnable<CompletableFuture<Either<Chunk, ChunkHolder.Unloaded>>> cir) {
```
Gegy: This would run for every chunk status- should probably filter to just do it on one status. Possibly the one that comes before `SPAWN`

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
```diff
@@ -0,0 +1,37 @@
...
+    public static void addSlimeChunk(ChunkPos chunk) {
+        slimeChunks.add(chunk);
+    }
+
+    public static void unloadSlimeChunk(ChunkPos chunk) {
+        slimeChunks.remove(chunk);
+    }
+
+    public static void clearChunks(ServerWorld world) {
+        new Thread() {
```
Gegy: Does this need to be on another thread?

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
```diff
@@ -0,0 +1,37 @@
+package me.jellysquid.mods.lithium.common.chunk;
+
+import java.lang.Thread;
+import java.util.HashSet;
+
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.util.math.ChunkPos;
+ 
+public class SlimeChunkStorage{
```
Gegy: Might be better to store this as a field on the Chunk rather

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkSerializerMixin.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.util.math.ChunkPos;
+import net.minecraft.world.ChunkSerializer;
+import net.minecraft.world.chunk.Chunk;
+import net.minecraft.world.poi.PointOfInterestStorage;
+import net.minecraft.server.world.ChunkHolder;
+
+@Mixin(ChunkSerializer.class)
+public class ChunkSerializerMixin {
+    @Inject(at = @At("RETURN"), method = "deserialize")
+    private static void cacheSlimeChunk(ServerWorld world, StructureManager structureManager, PointOfInterestStorage poiStorage, ChunkPos pos, CompoundTag tag, CallbackInfo ci) {
```
Gegy: Could probably avoid serialization of this altogether by recomputing in the ChunkStatus load task

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkSerializerMixin.java
```diff
@@ -0,0 +1,36 @@
...
+
+        if(isSlime) {
+            SlimeChunkStorage.addSlimeChunk(pos);
+        }
+    }
+
+    @Inject(at = @At("RETURN"), method = "serialize", cancellable = true)
+    private static void attachSlimeChunk(ServerWorld world, Chunk chunk, CallbackInfoReturnable<CompoundTag> cir) {
+        cir.getReturnValue().putBoolean("slimeChunk-LITHIUM", SlimeChunkStorage.checkSlimeChunk(chunk.getPos()));
+        SlimeChunkStorage.clearChunks(world);
```
Gegy: Will this cause a full clear after saving each chunk? Could this lead to chunks with missing data?

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkStatusMixin.java
```diff
@@ -0,0 +1,42 @@
...
+        if(world.getDimension() == DimensionType.getOverworldDimensionType()) {
+            for(Chunk currentChunk : chunks) {
```
Gegy: This should only operate on `chunks.get(chunks.size() / 2)` - the rest are context chunks :)

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
eatmyvenom: Issue with this is that it is done each time a chunk is serialized because idk where else to put it so it will nuke performance iterating through each slime chunk in the world every time a chunk is saved.

Gegy: Yeah, I think the solution of storing it as a field on the chunk would work much better and avoid this problem altogether

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
eatmyvenom: That was the initial idea but the slime spawning doesnt get sent a chunk so I cant really do anything about that because there is no good way of getting the chunk in ram from the blockPos because it would mean i have to deserialize it again each spawn

Gegy:
Also, what sort of performance hit is there from slime spawning normally?

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkSerializerMixin.java
eatmyvenom: i just fixed this part

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkStatusMixin.java
eatmyvenom: oh epic thx

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
Gegy: Spawning would only happen in loaded chunks, though. It should be fine to get the chunk from the world

eatmyvenom:
Well considering that any position can go through the full range of checks for slimes including making a new random and everything for any chunk in the world, i can imagine that whatever it is could be reduced quite a bit

jaskarth:
Is there any profiling data you can share here? I've never seen this show up on any profiler graph and am curious as to how much of a difference it makes.

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkSerializerMixin.java
eatmyvenom: I don't see why serialization here is a bad thing other than it means you don't have to make sure it has been serialized before.

src/main/java/me/jellysquid/mods/lithium/common/chunk/SlimeChunkStorage.java
eatmyvenom: Okay yea after some work this is completely gone in favor of your idea. (It doesnt work but still thats how its gonna be done)

src/main/java/me/jellysquid/mods/lithium/mixin/entity/slimechunk_cache/ChunkStatusMixin.java
eatmyvenom: I didn't see a nice way of doing this the way you mentioned.

Earthcomputer:
Agree with SuperCoder here, Random is very fast, I could imagine the hashing algorithm used for the caching actually being slower than the random number generation used to test if a chunk is a slime chunk. I could also imagine the rest of the mob spawning algorithm taking up a much more significant amount of the time. Do you have any profiling data to back this change up?

eatmyvenom:
Well I haven't been home in about a week and am on my laptop so it has been difficult to profile. I added a "slime spawing" to the vanilla profiler and did small measurements. That is promising but it's no real test. Today I will try to devise a good test and send the results.

eatmyvenom:
Also I would like to point out that the way this is implimented has changed drastically at this point. I should edit my PR message to reflect this.

ontrigger:
I did a little bit of profiling on my own and found that slime spawning does not contribute anything to the graph.
I was barely able to get it to even show up, so I doubt these perf optimizations are needed

eatmyvenom:
Well I did some (not great) profiling as well a few days ago and did notice that while this is small it was clearly non-zero which leads me to wonder what your tests were. Do you mind sharing the actual outputs as well as what your whole test setup was?

ontrigger:
Sure. I set up a 3x3 chunk perimeter with 4 3-block high platforms to spawn slimes in. 

I then made slimes spawn everywhere regardless of the 1 / 10 chance while keeping the random code running, like this:
spawntests2


Then, a command block kills all slimes every second.
Here are the results:
spawntests

Earthcomputer:
SlimeEntity doesn't even show up in the profile results. SpawnHelper does, which is what I told eatmyvenom 5 days ago. And even then it's 0.3% of the total time. This is a useless micro-optimization at best and makes slime spawn checking several times slower at worst.

Earthcomputer:
I want to see before and after profile results before I'm even slightly convinced.

eatmyvenom:
I added slime spawning specifically to the profiler to do my testing. As for it being useless, I don't really see a point in arguing over it. I agree its not really important but its up to 2no2name or jellysquid to decide if this doesn't belong in lithium at all.

2No2Name:
Neither me or Jelly want to add mixins that are likely not improving performance. All mixins take time to maintain for future versions, and therefore we want the mixins we have to be effective in improving performance. If a piece of code looks suspiciously unoptimized, make sure to check that it actually takes some runtime. In general it is a good idea to first profile the game (in whatever situation your world is realistically in) and find hotspots that can be optimized. Even if you were to speed up a method by 90%, it doesn't matter if it wasn't using any runtime anyways. 
If you come up with a somewhat realistic test case (e.g. big farm, even with /kill command), I will consider adding this.

eatmyvenom:
So as earth computer clearly stated, it really is quite a small difference, I have [uploaded the outputs](https://gist.github.com/eatmyvenom/d58793345fa24b0c92e302d97e1c171c) from the games internal profiler. The summary is that with the cache, the slime spawning was 1.00% of the time, and without the cache it is 1.26%. That makes this pr 0.26% faster than vanilla slime spawning for a single player. My test here was a superflat world with 4 128x128 platforms of jack'o'lanterns spaced apart 4 blocks on the y axis to give many spawn attempts.

eatmyvenom:
As for the state of this pr, it really does not add much to the game so I am going to give an attempt at one more optimization to this before closing it for it being really not worth anyones time.

eatmyvenom:
Closed because I could only get another 0.03% out of it

PR #96, opened 2020-08-01
change: Work on existing patches to squeeze out additional performance

maityyy:


2No2Name:
Did you notice any performance improvement with these changes?

maityyy:
Yes. These changes improve the performance of Villagers and Piglins by replacing their AI task code with a simpler one, eliminating the use of lambdas and Stream API. Also, this PR expands on several other patches that increase the overall performance in the normal world.

src/main/java/me/jellysquid/mods/lithium/mixin/world/chunk_inline_block_access/WorldChunkMixin.java
```diff
@@ -47,10 +49,19 @@ public BlockState getBlockState(BlockPos pos) {
...
     /**
      * @reason Reduce method size to help the JVM inline
-     * @author JellySquid
+     * @author Maity
      */
     @Overwrite
-    public FluidState getFluidState(BlockPos pos) {
-        return this.getBlockState(pos).getFluidState();
+    public FluidState getFluidState(int x, int y, int z) {
+        // WorldChunk#getFluidState(BlockPos) redirects to this method
```
2No2Name, line 56: What is the improvement here? The order in which methods are called is a bit different, and marginally speeding up the case of getting blocks outside the world does not seem necessary unless it is actually an often taken branch.

maityyy: > What is the improvement here? The order in which methods are called is a bit different, and marginally speeding up the case of getting blocks outside the world does not seem necessary unless it is actually an often taken branch.

`getFluidState(BlockPos)` redirects to `getFluidState(int, int, int)` when Lithium redirects it to the optimized version, but `getFluidState(int, int, int)` remained non-optimized. This change will optimize all methods.

P. S. I hope you understand, because I am from another country and I can make stupid mistakes in English.

2No2Name: The only place where `getFluidState(int, int, int)` is called is `getFluidState(BlockPos)` (and maybe some mods also call it)

maityyy: > The only place where `getFluidState(int, int, int)` is called is `getFluidState(BlockPos)` (and maybe some mods also call it)

Well, that's not bad, is it?

2No2Name:
I noticed some problems with your mixins and fixed them. See here: https://github.com/jellysquid3/lithium-fabric/commit/b162eea816bd620a2c4cb9eb96670eca6ce9f1cc or here: https://github.com/2No2Name/lithium-fabric/tree/maityyy-1.16.x/dev

maityyy:
Are these all problems?

2No2Name:
As there were so many problems I don't believe I caught all of them

maityyy:
I'll open it later, but without any problems.

maityyy:
Again.

PR #98, opened 2020-08-04
Fix explosions being different from vanilla

2No2Name:
Vanilla changed how explosions work: Air blocks now have no effect on the blast rays, and each explosion has its own context that can adjust the explosion resistance of blocks or decide if a block can be destroyed at all (Respawn anchor and entities have their own special context)


src/main/java/me/jellysquid/mods/lithium/common/util/TypeCast.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Class to work around being unable to cast SomeClassMixin to SomeClass inside SomeClassMixin.
+ * Add methods for ohter types when needed.
+ *
+ * Hopefully(!) the casts can be inlined and completely eliminated (casting SomeClass to SomeClass never doing anything).
+ *
+ * @author 2No2Name
+ */
+public class TypeCast {
+    public static Explosion toExplosion(Object returnValue) {
```
Earthcomputer: Why not just inline this cast?

2No2Name: I usually get problems when trying to cast to the mixined class. Either intellij, mcdev or the compiler complains, or it crashes when applying the mixin

2No2Name: inconvertible types, because ExplosionMixin doesn't extend Explosion.
This can be circumvented by casting like this: (Explosion)(Object)this
But I had problems with that approach in the past, but I don't recall when exactly they occur.

Kroppeb: Why is it cast to the Mixin in the first place? I thought you should cast it to an interface which the mixin implements?

Edit: nvm that wouldn't work for passing it as an argument, would it?
Can the mixin `extend` the class it's mixing in to? I can't recall.

Kroppeb: Even then, I would have made it a private function in the mixin, instead of making a static class just for this.

src/main/java/me/jellysquid/mods/lithium/mixin/world/explosions/ExplosionMixin.java
```diff
@@ -191,20 +198,25 @@ private void performRayCast(Random random, double vecX, double vecY, double vecZ
     }
 
     /**
-     * Called for every step made by a ray being cast by an explosion.
+     * Called for every step made by a ray being cast by an explosion. fai
```
jaskarth: what does `fai` mean here? is it a typo? 😅

Earthcomputer: ferks as intended

2No2Name: probably me just typing by accident

jaskarth (approved):

src/main/java/me/jellysquid/mods/lithium/mixin/world/explosions/ExplosionMixin.java
```diff
@@ -220,44 +232,47 @@ private float traverseBlock(float strength, int blockX, int blockY, int blockZ,
...
+                    // If the block state is air, it cannot have fluid or any kind of resistance, so just leave
+                    if (blockState.getBlock() != Blocks.AIR) {
+                        // Rather than query the fluid state from the container as we just did with the block state, we can
+                        // simply ask the block state we retrieved what fluid it has. This is exactly what the call would
+                        // do anyways, except that it would have to retrieve the block state a second time, adding overhead.
+                        FluidState fluidState = blockState.getFluidState();
+
+                        // Get the explosion resistance like vanilla
+                        blastResistance = this.behavior.getBlastResistance((Explosion) (Object) this, this.world, pos, blockState, fluidState);
+                        break labelGetBlastResistance;
```
Gegy, line 260: Possibly cleaner to have this be a separate function to avoid the label and break

PR #102, opened 2020-08-17
new: Optimize block entity ticking & unloading

maityyy:
This PR introduces a new optimization patch that uses hash tables as a specifically optimized collection option for the list of unloaded block entities in the world, which *greatly improves performance* when unloading large numbers of them (also see [MC-117075](https://bugs.mojang.com/browse/MC-117075)).

Based on my tests, it took about *~60%* less time to unload a full chunk of chests than vanilla. However, my computer is *too weak* and the numbers may differ significantly from mine. But what I can say for sure is that it will definitely improve performance.

**P. S.** I am from another country and English is not my first language. If I made stupid mistakes, please correct me. I'll be *very* grateful.


src/main/java/me/jellysquid/mods/lithium/common/util/collections/HashSetBackedList.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Override
+    public Stream<E> parallelStream() {
+        return this.backedCollection.parallelStream();
+    }
+
+    @Override
+    public boolean addAll(int index, Collection<? extends E> c) {
+        return false;
```
2No2Name: If any other mod decides to use this method, we shoudn't fail silently if we don't implement it.
I doubt many other mods mess with unloading block entities. Using HashedList might be a good option still.
Otherwise we need to throw UnsupportedOperationExceptions as failing silently can lead to a lot more problems

2No2Name (changes requested):
LGTM besides the wrapper that does not implement all methods properly

src/main/java/me/jellysquid/mods/lithium/common/util/collections/HashSetBackedList.java
maityyy: I decided not to use `HashedList` as it uses two collections at once. These are more resources. Of course, this would also speed up the process of unloading block entities, but then the patch was not effective.

src/main/java/me/jellysquid/mods/lithium/common/util/collections/HashedList.java
```diff
@@ -16,6 +16,11 @@ private HashedList(List<T> list, Set<T> set) {
         this.set = set;
     }
 
+    public HashedList(List<T> list) {
+        this.list = list;
+        this.set = new HashSet<>();
```
2No2Name: this.set.addAll(list) is missing here. You are expecting the list to be empty, but others who use the function in the future might not expect that.

2No2Name (approved):

PR #103, opened 2020-08-18
update: Completely exclude Stream API from mob goal selection system

maityyy:
This small PR completely excludes the *Stream API* with lambdas from the `GoalSelector` class (`remove` method), which improves the game performance **with large counts of some mobs** (skeletons, zombies, cats and ocelots).

The changes in mspt are very difficult to see. I have used standard debugging using **F3+Alt** and have always noticed small performance gains. No problems were found in the behavior of the mobs.

**P. S.** Also, if I have made any stupid mistakes, please correct me. **English is not my native language.**

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/ai/goal/GoalSelectorMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/ai/goal/GoalSelectorMixin.java
@@ -11,6 +11,7 @@ import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import java.util.EnumSet;
+import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


maityyy:
I've updated the description to make it clearer what this PR does. Sorry for my mistakes.

2No2Name (approved):
The code you are changing does not look like it is called often. The calls from cats/ocelots and skeleton horses are very rare. However Skeletons and Zombies being initialized might be relevant for some fast mob farms. I doubt that whatever improvement you noticed was from this patch though.

maityyy:
Yes, it is rarely called. I also expect other mods to use it. I think we shouldn't leave it un-optimized when the whole goal selector has been rewritten. The function itself used slightly less memory in tests and was faster.

maityyy:
It doesn't change anything on the Java versions above.

PR #109, opened 2020-08-23
General code and performance improvements

deeprobin:


src/main/java/me/jellysquid/mods/lithium/mixin/ai/poi/fast_init/PointOfInterestStorageMixin.java
```diff
@@ -35,22 +35,20 @@ public PointOfInterestStorageMixin(File directory, Function<Runnable, Codec<Poin
      */
     @Overwrite
     public void initForPalette(ChunkPos chunkPos_1, ChunkSection section) {
-        ChunkSectionPos sectionPos = ChunkSectionPos.from(chunkPos_1, section.getYOffset() >> 4);
-
+        final ChunkSectionPos sectionPos = ChunkSectionPos.from(chunkPos_1, section.getYOffset() >> 4);
         PointOfInterestSet set = this.get(sectionPos.asLong()).orElse(null);
 
-        if (set != null) {
+        if(set == null) {
```
maityyy: Space is missing here.

src/main/java/me/jellysquid/mods/lithium/common/util/streams/ConcatSpliterator.java
```diff
@@ -0,0 +1,115 @@
...
+
+                    & ~(Spliterator.DISTINCT | Spliterator.SORTED
+                    | (unsized ? Spliterator.SIZED | Spliterator.SUBSIZED : 0)))
+
+                    : bSpliterator.characteristics();
+        }
+
+        @Override
+        public Comparator<? super T> getComparator() {
+            if (beforeSplit)
```
maityyy: {} Is missing here.

maityyy (changes requested):

maityyy:
Are all these `final` modifiers needed? As I recall, sometimes this is a signal to the JVM to optimize the code and this can really improve performance, but the JVM already knows what needs to be optimized in most cases.

deeprobin:
> Are all these `final` modifiers needed? As I recall, sometimes this is a signal to the JVM to optimize the code and this can really improve performance, but the JVM already knows what needs to be optimized in most cases.

I personally feel that this is necessary. Of course, the JVM would probably notice this itself in many places, but the `final' modifiers don't harm the code performance-wise, and possibly they support the JVM. It also improves code readability.

[DZone: Final Keyword and JVM Memory Impact](https://dzone.com/articles/final-keyword-and-jvm-memory-impact)

2No2Name:
> In most common cases are `<`-operators faster than `<=`-operators.

Do you have any source for this claim?

2No2Name:
Do you have any measurements by how much the performance is improved by this PR?

src/main/java/me/jellysquid/mods/lithium/common/util/streams/ConcatSpliterator.java
```diff
@@ -0,0 +1,117 @@
+package me.jellysquid.mods.lithium.common.util.streams;
+
+import java.util.Comparator;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+
+abstract class ConcatSpliterator<T, S extends Spliterator<T>>
```
2No2Name, line 7: This code is taken from a library with a license that is incompatible with ours, isn't it?

jellysquid3:
The changes introduced by this PR are superficial and look to be done entirely by automated tooling. Defining every variable as `final` (especially local vars) and using ternaries everywhere will not help the JVM produce faster code and only increase code verbosity.

We are not interested in PRs which only perform code style changes, especially those that make such large sweeping changes like this one with zero justification or prior communication.

src/main/java/me/jellysquid/mods/lithium/common/util/streams/ConcatSpliterator.java
deeprobin: No this is slightly modified code from OpenJDK which is licensed under the GPL license and is therefore compatible with the LGPL.

skyrising: OpenJDK is licensed under GPLv2 which is incompatible with this projects LGPLv3: https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

PR #110, opened 2020-08-25
new: Speeding up common BlockPos functions

maityyy:
This PR will improve some of the game's processes (chunk ticking, structure and feature generation) by simply speeding up common `BlockPos` functions like `up`, `down`, `west`, etc.

All these functions are redirected to offset, where the calculation takes place using `Direction`, however, we can manually write the logic of these functions, avoiding these overheads, thereby speeding up them.

According to my tests, in the ordinary world, tree generation has become **~10-15%** faster, **but** I think it should be much better, it's just that my computer is too bad for such tests.

**P. S.** I am from another country and English is not my first language. Sorry for the silly mistakes. Correct me!


maityyy:
I tested with a separate mod and completely forgot that import of javax annotations is required for some reason. I'll fix it now.

Kroppeb:
Huh, i assumed that once those methods are warmed up this shouldn't really change anything

maityyy:
> Huh, i assumed that once those methods are warmed up this shouldn't really change anything
I don't really understand what you mean.

build.gradle
```diff
@@ -26,10 +26,13 @@ minecraft {
 }
 
 dependencies {
-	//to change the versions see the gradle.properties file
+	// To change the versions see the gradle.properties file
 	minecraft "com.mojang:minecraft:${project.minecraft_version}"
 	mappings "net.fabricmc:yarn:${project.yarn_mappings}:v2"
 	modCompile "net.fabricmc:fabric-loader:${project.loader_version}"
+
```
2No2Name, line 33: It doesn't look like the changes in this file are necessary for your commit at all. Where do you need the javax annotations?

2No2Name (approved):
Looks good to me. I did not check whether all the offsets are added on the correct axis yet. If someone wants to do that, comment when you did

build.gradle
maityyy: idk, maybe there is some bug. These annotations are required at compilation and if they are not there will crash.

maityyy:
> Looks good to me. I did not check whether all the offsets are added on the correct axis yet. If someone wants to do that, comment when you did

I checked and everything was ok.

jellysquid3:
The jsr305 dependency is required because of an issue with Mixins. We're fine to add it here as a compile-only dependency.

i509VCB:
Actually I had this issue with game rule api in fabric api. I think this was specifically a javac error.

2No2Name:
From my testing there is no speedup.

2No2Name:
I am closing this PR as we cannot expect any speedup from it. (If you think this is wrong, please provide such a benchmark. You said that tree generation is faster, but you didn't show or explain how to test that.)

PR #113, opened 2020-08-26
change: Initially enable the palette container lock removal patch

maityyy:
As I wrote in the [discord](https://discordapp.com/channels/602796788608401408/648967810139029505/748258210128986132), it is absolutely safe. We are waiting for players' reports.


maityyy:
Just as a reminder, all fixes are coming later.

2No2Name:
Why are you replacing Jelly's original implementation that overwrote the lock and unlock methods?

maityyy:
I did this following the example of optimizing entity DataTracker so that the JVM sees empty methods and decides to completely remove this field.

jellysquid3:
Implemented with 656ef5e9.

PR #122, opened 2020-08-31
new: Remove some streams from brain tasks

maityyy:
GC doesn't really like the Stream API in these places.


maityyy:
I also have a list of brain tasks, where there are still streams, but not everything is possible and need to replace them with simple iteration:
```
TimeLimitedTask.java
OpenDoorsTask.java
VillagerBreedTask.java
WalkHomeTask.java
WanderIndoorsTask.java
WorkStationCompetitionTask.java
TakeJobSiteTask.java
```

jellysquid3:
Do you have real-world (or even synthetic) benchmarks for these changes? These patches copy a tremendous amount of logic, and while it's easy to point fingers at the Stream API for being slow and difficult for the JVM to optimize, it's not always the case.

maityyy:
> Do you have real-world (or even synthetic) benchmarks for these changes? These patches copy a tremendous amount of logic, and while it's easy to point fingers at the Stream API for being slow and difficult for the JVM to optimize, it's not always the case.

Minus ~12 mspt for me.

2No2Name:
In which setting?

maityyy:
> In which setting?

Vanilla + latest build of Lithium with and without my changes (this PR).

maityyy:
I did not find any bugs, this PR for me subtracts about 10-20 mspt with ~40 villagers.

2No2Name:
I cannot confirm those measurements. I see no difference (before and after: around 8-9mspt) in a simple world with ~ 100 villagers in a village. Can you provide some more details on how I can reproduce your mspt improvements?

PR #123, opened 2020-08-31
new: optimize `BlockPos.iterateOutwards` by caching offsets

2No2Name:
This PR implements the optimization from https://github.com/jellysquid3/lithium-fabric/pull/90 in a more generalized way, so that all calls to BlockPos.iterateOutwards can use cached values ( there are 31 different possible range parameters in vanilla only).
Also this PR tries to implement the cache in a threadsafe manner.

This is currently a draft, because profiling still needs to be done.


maityyy:
Changes to code style: everything that is in `common.cached_block_pos` should be moved to `common.util`/`common.util.collections`. There is a space before start of the comment, and word must start with a capital letter.

maityyy:
Oh, and the optimization patch itself, I think, should be referred to `util.fast_math`.

2No2Name:
I don't think it makes sense to merge this. The worldgen code likely does not depend on the iteration order. In this case reordering the blockstate accesses makes sense. For Monsters searching for the nearest block, the block counting system could be used to cache the result until the monster moves or a block of the type appears or disappears.

PR #130, opened 2020-09-05
Apply thread-local cache to InitLayer and MergingLayer

Gegy:
Previously, we were only applying this to `ParentedLayer`, which would eliminate the majority of potential race conditions. This would make any race condition issue practically impossible in a vanilla context, but when adding more threads, it becomes much more likely to occur through the alternative `InitLayer` and `MergingLayer`, which were missed by accident in the original PR.


jaskarth (approved):

2No2Name:
Looking at the code now. Are you saying the vanilla code has race conditions?

src/main/java/me/jellysquid/mods/lithium/common/world/layer/CachedLocalLayerFactory.java
```diff
@@ -1,42 +1,56 @@
...
+            }, sampler1, sampler2);
+        });
+    }
 
-        return context.createSampler((x, z) -> {
-            context.initSeed(x, z);
-            return layer.sample(context, parentSampler, x, z);
-        }, parentSampler);
+    private static <R extends LayerSampler> LayerFactory<R> createMemoized(LayerFactory<R> factory) {
+        ThreadLocal<R> cached = new ThreadLocal<>();
```
2No2Name: Is this different from `return ThreadLocal.withInitial(factory::make)::get;`?

2No2Name (approved):
In general the code looks good to me. I wonder how you chose the cache capacities, as they are different from vanilla, but you are not sharing the cache between threads.

src/main/java/me/jellysquid/mods/lithium/common/world/layer/CachedLocalLayerFactory.java
Gegy: That should work

Gegy:
I don't remember exactly how we ended up with these cache sizes- I recall there was the consideration of how many threads there are, but also the reduced memory overhead with this implementation in comparison to vanilla.

PR #145, opened 2020-09-23
 new: use precomputed merged retracting piston base shapes and avoid VoxelShapes.union on empty shapes

2No2Name:
As @jellysquid3 said on discord, getting the collision/outline shapes from moving blocks is quite an expensive call, because VoxelShapes.union is used.
Vanilla calls VoxelShapes.union for joining the piston base shape with the piston head shape for retracing pistons. As there are only a few possible configurations ( 6 directions, 3 different extension progress values), this PR precomputes all of them.

In other cases, vanilla also calls VoxelShapes.union, but with the first argument being an empty shape. The effect of this method is that the 2nd argument is "simplified" (newly constructed). We cannot skip the simplification, as e.g. a shifting floor of soul sand does not have hitboxes inside the moving blocks in vanilla, but it would have without the simplify() call. Therefore this PR uses ThreadLocal WeakHashMap's to cache the result of the simplify() call. As simplify() is only called after the shape was offset according to the moving block's progress, caching for all variants has to be done seperately.

When stepping though the debugger I noticed that the client is also calling this code very often

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/block/moving_block_shapes/PistonBlockEntityMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/block/moving_block_shapes/PistonBlockEntityMixin.java
@@ -1,0 +1,125 @@
+package me.jellysquid.mods.lithium.mixin.block.moving_block_shapes;
+
+import me.jellysquid.mods.lithium.common.shapes.OffsetVoxelShapeCache;
+import net.minecraft.block.BlockState;
+import net.minecraft.block.Blocks;
+import net.minecraft.block.PistonBlock;
+import net.minecraft.block.PistonHeadBlock;
+import net.minecraft.block.entity.PistonBlockEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.Direction;
+import net.minecraft.util.shape.VoxelShape;
+import net.minecraft.util.shape.VoxelShapes;
+import net.minecraft.world.BlockView;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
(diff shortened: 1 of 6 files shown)


src/main/java/me/jellysquid/mods/lithium/mixin/block/moving_block_shapes/PistonBlockEntityMixin.java
```diff
@@ -0,0 +1,129 @@
...
+                for (float offset : offsets) {
+                    VoxelShape offsetHead = headShape.offset(facing.getOffsetX() * offset,
+                            facing.getOffsetY() * offset,
+                            facing.getOffsetZ() * offset);
+                    mergedShapes[getIndex(isShort, facing.getId(), offset)] = VoxelShapes.union(baseShape, offsetHead);
+                }
+            }
+        }
+
+        for (VoxelShape shape : mergedShapes) {
```
maityyy: I think this iteration wastes CPU resources. Isn't it possible to move it after initialization to line 64?

2No2Name: This iteration is just a sanity check. It would just throw an Exception if the getIndex function is somehow incorrect. As it is only run once per start of the game, I don't think it is important to avoid the loop, but instead of moving it, it can just be removed

2No2Name:
Here is a setup that demonstrates the best-case for the simplify cache (cauldron as complex shape, lots of armorstands trying to "simplify" it before colliding with it).

2No2Name:
I believe the concurrent hashmap lookup can be replaced by instead safely publishing the effectively final VoxelShapes in a field of their parent.

PR #148, opened 2020-09-29
new: optimize VoxelShapes.matchesAnywhere using simplicity of cuboid shapes

2No2Name:
This PR avoids the allocations in VoxelShapes.matchesAnywhere (PairLists and lambdas) in cases where it is called with at least one cuboid shape as argument. If both arguments are cuboid shapes the collision calculation is very simple.
The correctness was tested by randomly trying out millions of collisions.
Performance testing still needs to be done


2No2Name:
Profiling showed about 50% lower runtime in a test setup that contains both collisions between two cuboids, and between cuboids and more complex shapes. (Setup: took the Test class and only ran the vanilla, and then only ran the modded code (with vanilla fallback), result: runtime is 50% lower when running modded code)

PR #152, opened 2020-10-13
change: allow compatibility with both 1.16.2 and 1.16.3

LucilleTea:
The differences between 1.16.2 and 1.16.3 are minor and don't interact with Lithium.
It seems that people are continuing to use 1.16.2, so it'd be nice to allow future versions of Lithium to run there. Otherwise people are forced to use an outdated version or edit the .jar manually.


jellysquid3:
LGTM. This is generally an exceptional case, but for this version bump alone it's safe.

PR #154, opened 2020-10-17
Add support for 1.16.4

Obydux:


2No2Name:
We added the breaks clause to prevent users from loading the mod in the wrong versions. As 1.16.4 is not released yet, we can't know whether this version is compatible.

PR #178, opened 2020-12-26
Implement a faster biome array implementation for the overworld

jaskarth:
With the introduction of vertical biome support in 1.15, biome storage has moved from a 16 * 16 array in favor of a 4 * 64 * 4 low-resolution array where each element represents a 4x4x4 cube which is then upscaled to represent the whole chunk. While the nether and end use 3d upscaling via `VoronoiBiomeAccessType`, the overworld uses 2d upscaling through `HorizontalVoronoiBiomeAccessType`, so we can get away with only sampling the horizontal coords and using it to fill in the rest of the column. A quick sampling shows a **2x** improvement in biome array setting times. Granted, it wasn't much to begin with in vanilla, but with more complicated biome layer setups this can make a huge difference.
Before:
image
After:
image

I would like to thank @alcatrazEscapee for the original implementation of this patch and allowing the usage of his code :)

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_biome_array/ChunkStatusMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_biome_array/ChunkStatusMixin.java
@@ -1,0 +1,28 @@
+package me.jellysquid.mods.lithium.mixin.gen.fast_biome_array;
+
+import java.util.List;
+
+import me.jellysquid.mods.lithium.common.world.biome.HorizontalBiomeArray;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.util.registry.Registry;
+import net.minecraft.world.biome.source.HorizontalVoronoiBiomeAccessType;
+import net.minecraft.world.chunk.Chunk;
+import net.minecraft.world.chunk.ChunkStatus;
+import net.minecraft.world.chunk.ProtoChunk;
+import net.minecraft.world.gen.chunk.ChunkGenerator;
+
+@Mixin(ChunkStatus.class)
+public class ChunkStatusMixin {
+    @Inject(method = "method_16570", at = @At("HEAD"), remap = false, cancellable = true)
+    private static void populateBiomes(ServerWorld world, ChunkGenerator generator, List<Chunk> surroundingChunks, Chunk chunk, CallbackInfo info) {
+        if (world.getDimension().getBiomeAccessType() instanceof HorizontalVoronoiBiomeAccessType) {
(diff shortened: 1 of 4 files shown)


DragonEggBedrockBreaking:
You forgot to add the Mixin to the config file. I PR'd to your fork so you can pull in and this PR will be updated. I don't care about contributing, so you can manually type the Mixin into config rather than pulling in my config file (if you want).

jaskarth:
Ended up doing it manually since it's just one line. Thanks though!

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_biome_array/ChunkStatusMixin.java
```diff
@@ -0,0 +1,36 @@
...
+public class ChunkStatusMixin {
+
+    /**
+     * We inject into the lambda in the BIOME status to construct a horizontal biome array if the biome access type is horizontal.
+     * Since the method is a lambda, it's not mapped and the unresolved mixin warning can be ignored.
+     *
+     * @author SuperCoder79
+     */
+    @SuppressWarnings("UnresolvedMixinReference")
+    @Inject(method = "method_16570", at = @At("HEAD"), remap = false, cancellable = true)
```
DragonEggBedrockBreaking, line 28: Not a programmer, but this seems to be related to #205 and a94e810 and may/may not require a (small?) code edit

maityyy: The multiline is only needed if the line is very long

src/main/java/me/jellysquid/mods/lithium/mixin/gen/fast_biome_array/ChunkStatusMixin.java
```diff
@@ -0,0 +1,36 @@
...
+@Mixin(ChunkStatus.class)
+public class ChunkStatusMixin {
+
+    /**
+     * We inject into the lambda in the BIOME status to construct a horizontal biome array if the biome access type is horizontal.
+     * Since the method is a lambda, it's not mapped and the unresolved mixin warning can be ignored.
+     *
+     * @author SuperCoder79
+     */
+    @SuppressWarnings("UnresolvedMixinReference")
```
TheGlitch76, line 27: This should be replaced by `@Dynamic("lambda in <method>)`
It's the official way to inject into something the Mixin AP can't see

2No2Name:
I am considering to merge this in the 1.17 branch. Is the horizontal biome array only meant to be used during generation, or also after a chunk reload? Either way the save format needs to be like vanilla, as users may uninstall lithium or disable the mixin at any time. Could this be an issue?

alcatrazEscapee:
The horizontal biome array is just a modification of one of the constructors. It only affects sampling when the initial biome layers are passed in.

- Save format is identical, as once constructed, the horizontal biome array behaves (and stores the exact same data) that vanilla does.
- It will not be used after a chunk reload (nor does it gain any benefit from doing so) as all biomes are saved to the chunk and do not gain any benefit from the horizontal only sampling that this biome array uses.

jaskarth:
Closing as the changeset is out of date and doesn't apply to modern versions.

PR #179, opened 2020-12-27
Implement a cache for chunk generator settings

jaskarth:
Chunk generator settings are passed to the noise chunk generator through a supplier, which retrieves a given chunk generator settings registry key from the registry. The problem arises as this setting is retrieved from the supplier to get the sea level, which is called **every single time a block is placed by the generator, for a total of 65536 registry lookups per chunk.** As you can imagine this isn't all that fast and removing it speeds up the chunk generation time by a moderate amount (13% of CPU time spent in populateNoise without the patch vs 10% with the patch). My solution to this was to stick the supplier into a Lazy<> so it's computed once and then cached, but I'm open to suggestions. (The sea level can't be computed ahead of time due to a quirk in the new registry system, which would have been the best option here.)

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/gen/cached_generator_settings/NoiseChunkGeneratorMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/gen/cached_generator_settings/NoiseChunkGeneratorMixin.java
@@ -2,8 +2,10 @@ package me.jellysquid.mods.lithium.mixin.gen.cached_generator_settings;
 
 import java.util.function.Supplier;
 
+import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.Overwrite;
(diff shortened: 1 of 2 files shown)


Gegy:
It might make sense to lazily cache the sea level itself by handling the lazy logic manually (assuming the sea level can never change?)

jaskarth:
Hmm, that could be another option- I just used Lazy because it handled everything automatically, but we can look into handling it manually if it helps performance

2No2Name:
Looks good to me. Have you done any profiling after the last commit?

2No2Name:
Is there any reason to not initialize the cached sea level field in the constructor?

maityyy:
You forgot config entry https://github.com/jellysquid3/lithium-fabric/blob/ea97a63a3351b6f706e03a58e8ccf9ca749ddca7/src/main/java/me/jellysquid/mods/lithium/common/config/LithiumConfig.java#L75

2No2Name:
I moved the initialization to the constructor and added the mixin rule. Squashed / merged manually

PR #180, opened 2020-12-27
Extend ChunkRegionMixin to use the chunk array for setBlockState and getTopY

jaskarth:
This PR adds to ChunkRegionMixin by using the faster chunk array for the setBlockState and getTopY methods to avoid pointer dereferencing and making them easier to inline, like the existing getBlockState method. With some profiling this speeds up worldgen time by about 10% compared to without the patch. My one concern with this change is should the chunk array access be extracted to another method as the code is reused a few times- I was going to do that but I was unsure how it would affect the said goal of making the method easier to inline.


2No2Name:
edited and merged after discussion on discord

PR #186, opened 2021-01-09
Hopper Item Search optimisation

28Smiles:
The hoppers collision shape already prohibits items to reach the occluded shape. Reduces entity search overhead by 50% and removes the need to merge lists.

I am pretty sure but not entirely sure this does not break things. There are Possibilitys to detect it with command blocks. But normal vanilla experience should not change, since item Teleportation is not a thing.

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/block/hopper/HopperBlockEntityMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/block/hopper/HopperBlockEntityMixin.java
@@ -1,0 +1,42 @@
+package me.jellysquid.mods.lithium.mixin.block.hopper;
+
+import net.minecraft.block.entity.Hopper;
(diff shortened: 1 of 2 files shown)


maityyy:
If this PR can break worlds with command blocks, then Lithium is unlikely to accept it...

28Smiles:
The scenario of a break (which means the hopper tryes to collect an item it would not in vanilla) is really niche and unlikely, there has to be an item teleported into the hopper block itself at a certain position. It is most likely unintentional to do so. I am not really concerned I might break something used by players.

2No2Name:
The change in behavior is detectable by contraptions in two ways: 

1. The pickup area is different (yes only slightly, but we aim to not have those differences)
2. The order in which items are picked up is different (yes most people don't care)

Also there already exists https://github.com/2No2Name/hopperOptimizations and I am planning to merge that in the future. (have to profile the features a bit and see if all of them provide enough value)

PR #214, opened 2021-02-13
change: remove seemingly useless part of lithium.mixins.json

DragonEggBedrockBreaking:
I asked in the CaffeineMC discord (#lithium), and I was told that it was useless, so I made this pull request.

### DIFF
--- a/src/main/resources/lithium.mixins.json
+++ b/src/main/resources/lithium.mixins.json
@@ -6,8 +6,6 @@
   "injectors": {
     "defaultRequire": 1
   },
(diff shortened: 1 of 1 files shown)


TheGlitch76:
this is pointless, you may as well leave it there so that if someone adds a mixin to a client-only class they put it in the right spot

DragonEggBedrockBreaking:
> 
> 
> this is pointless, you may as well leave it there so that if someone adds a mixin to a client-only class they put it in the right spot

Ok.

PR #226, opened 2021-04-09
Updated loom to 0.6-SNAPSHOT (1.16)

krolik-exe:

PR #227, opened 2021-04-09
Updated loom to 0.6-SNAPSHOT (1.17)

krolik-exe:


DragonEggBedrockBreaking:
See my comment on #226, and I also think that if/when one of my previous PRs gets merged, the branch would be merged into 1.17, *if* the 1.17 branch continues very active development.

DragonEggBedrockBreaking:
Implemented into 1.17/dev, so this PR is not needed

PR #271, opened 2021-08-19
update gradle,fabric,lazydfu

backryun:


2No2Name:
I updated the gradle dependency and migrated to the latest yarn mappings now. 

I am not going to update lazydfu to a non-release version. If we experience any problems with the current release version that option will be considered. The hopper code change in this PR is probably just and accident and I therefore ignored it.

triphora:
> The hopper code change in this PR is probably just and accident and I therefore ignored it.

That was a mapping change, and also the reason that the CI build for the most recent commit failed.

2No2Name:
Interesting. I did run migratemappings, but it did not work for me. Thanks

2No2Name:
It seems like you did not run migrate mappings, because you did a different change than the tool. You can run migratemappings with the command `gradlew migrateMappings --mappings "1.17.1+build.40"`. Only after it is done, change gradle.properties and move all files from the remappedSrc directory to the correct src subdirectory

backryun:
thanks for advice about it.

PR #287, opened 2021-10-28
Fix compiler warnings, update dependences, cache Gradle Dependecies, etc.

eshankiyer:


amnotbananaama:
Thank you for the pull request. While it is appreciated, I wanted to share some concerns:
- You are not very descriptive of what exactly your PR does. What compiler warnings does it fix? What dependencies are you updating and why? Answering questions like these yourself make it easier for reviewers to evaluate the PR.

- The number of commits alone, makes it difficult to understand at a glance what exactly your PR is. Please try in the future to keep PRs to a reasonable size.

- Additionally, you have many contradictory commits where you revert something reverting something else, etc etc.

- Changing JitPack, GitHub Actions, and GitHub release builds of Lithium to use Java 17 is IMO a non-starter at this time. We shouldn't *require* a newer version of Java than Mojang without clear reason, and that is likely to cause compatibility issues on user's ends.

- Hard-coding the Minecraft version in `build.gradle` rather than using the template string is IMO an unneeded change. It will just end up being one more thing to update.

Please don't view this comment as derogatory or attempting to belittle your efforts! Your efforts to improve Lithium are appreciated. I am just pointing these things out so that you are aware of them for the future.

eshankiyer:
First of all, sorry the amount of Git commits is so large, I am having some difficulty editing the files locally on my computer so editing on GitHub is all I have for now.

Secondly, as shown below this pull request fixes the following compiler/loom warnings
image

Thirdly, I did test this out with Java 16, both vanilla OpenJDK and various forks, and I found no such incompatibility.

 Additionally updating Loom and Yarn versions due to in my perspective, better speed and reliability in builds.

Lastly, hardcoding the Minecraft version in the `build.gradle` file was temporary, and I have now fixed it in my fork of this repository

I'll be reopening a different pull request with the same content and with a description.

PR #299, opened 2021-11-27
Update to 1.18-rc3 mappings

msfjarvis:
### Proposed Changes

- Updates to 1.18-rc3 mappings, which in turn required the following changes
  - Implemented `Palette#copy` for `LithiumHashPalette`.
  - Updated the method injection in `LivingEntityMixin` from `getEquipment` to `getEquipmentChanges`.
  - Updated `shapes.lazy_shape_context` to account for changes to the constructor and removal of the `boots` field.
  - Removed `entity.gravity_check_block_below` mixin, since the `calculatePushVelocity` method being overridden by the mixin appears to no longer exist since 1.18-rc1.

Updates to Gradle 7.3

Supersedes #295


2No2Name:
Manually merged

PR #305, opened 2021-12-07
fix: Make GH Actions consistent with Sodium's

Crackers0106:
Noticed while updating to the latest dev builds of Lithium and Sodium that the artifacts produced were different, such as Sodium's naming for the build artifacts file being `sodium-artifacts-(branch name)` and only including the Minecraft compatible jar file.

Ideally, it feels like these should be pretty much the same between the three mods, and could make it easier to tell which file has which artifacts inside if you have a dozen files named `build-artifacts.zip` downloaded with no way to tell what they're for without checking inside them like I do.

This is my first PR, so I hope I'm doing this right

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -4,26 +4,42 @@ on: [ push, pull_request ]
 
 jobs:
   build:
-      runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
 
-      steps:
-          -   uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


2No2Name:
# Filter built files to disregard -sources and -dev, and leave only the minecraft-compatible jars.
        path: build/libs/*[0-9].jar
Doesn't seem to work

2No2Name:
This clearly happens because lithium uses a different jar naming system. Changing to sodiums naming system might be a good idea

PR #319, opened 2021-12-19
skipping steps in voronoi using prediction

FxMorin:
The title says it all


Titaniumtown:
I've found no issues with these optimizations personally. Would love to see numbers on actual performance though!

Titaniumtown:
Why was this closed?

2No2Name:
Closed due to the contributor being banned from CaffeineMC projects

Titaniumtown:
Oh?

jellysquid3:
We believe that the user in question was involved in distributing targeted malware to other users outside of our projects. As a result, we don't feel confident about the intention of any patches they have submitted, nor do we want to risk accepting code that could act maliciously in an obfuscated manner.

Issue #318, opened 2021-12-19
Crash report: [Server thread/ERROR]: Encountered an unexpected exception java.lang.NullPointerException: Cannot invoke "net.minecraft.class_1408.method_6345()" because "registeredNavigation" is null


Nekkok:
### Version
> lithium-fabric-mc1.18.1-0.7.6

### Reproduction Steps
> Create new world
> Wander around in survival / creative mode from 1 minute to 15 minutes
> The game crashes and it logs me out from the server, even though I'm playing in single player (of course, it happens in multi too)

### Crash Report File
https://gist.github.com/kxkxrxtx/86c9b39608d3d8677b3e159d9df826c8

### Additional Information
Hi, I'm reporting this rare issue I'm having only when Lithium is installed. It's happening in Minecraft 1.18.1 / Fabric 0.12.12 and it was also happening in Minecraft 1.18 / Fabric 0.12.8. It's probably because another mod, but I can't tell what is the cause of it. Thanks.

FxMorin:
I can actually confirm this. I thought it was just a fluke, although I also got the issue in singleplayer where I got kicked to the server list screen. Without any errors, was even debugging at the time.
I was only running Lithium & Carpet

2No2Name:
Temporary workaround until the next release:
Add `mixin.entity.inactive_navigations=false` to the LithiumConfig

PR #321, opened 2021-12-19
Null path branch Crash Fix

FxMorin:
Fixed missing `isRegisteredToWorld()` call before `registeredNavigation.getCurrentPath()`


FxMorin:
Fixes issue #318

PR #327, opened 2021-12-21
new: Replace Math.Round with Faster Implementation

FxMorin:
The Jafama FastMath Library has an implementation of Round that is 1.1x faster than Java's Math Library while keeping 100% accuracy.

I made mixins to modify most implementations of Math.round to use the faster implementation, although the main performance gain is actually within `CuboidVoxelSet` & `VoxelShapesMixin` which is used for Lithium's hitbox & collision optimizations. Both of which still being one of the slower parts of the game.

This is a very small optimization


DragonEggBedrockBreaking:
But now doesn’t mixin.shapes.specialised_shapes rely on mixin.math.fast_round, and shouldn’t that be in config as a mixin dependency, like how mixin.block.hopper has a few rule dependencies at the bottom?

FxMorin:
@DragonEggBedrockBreaking FastMath is not a mixin, so specialised_shapes can still use it without the config option being set to true

src/main/java/me/jellysquid/mods/lithium/common/util/math/FastMath.java
```diff
@@ -2,42 +2,21 @@
...
 
-    private static final int MAX_FLOAT_EXPONENT = 127;
-    private static final int MAX_DOUBLE_EXPONENT = 1023;
-
-
     /**
      * @author FX - PR0CESS
-     * Using jafama FastMath Library which is 1.1x faster while keeping 100% accuracy
+     * Custom rounding method which is even faster while keeping 100% accuracy
+     * ~1.25x faster than @link{Math#round(float)}
```
maityyy: {@link Math#round(float)}

FxMorin: thanks

malte0811:
The behavior of the new implementation doesn't always match that of `Math.round`: The float version returns 1 for `a = 0.49999997f` (int bits: `1056964607`), when it should be returning 0 (`a+.5f` is exactly 1 due to floating point error). It's also wrong for "exact" negative half-integers (`-0.5f` etc), where `Math.round` rounds up while the new implementation rounds down; and odd integers the range where the resolution of floats is 1 (e.g. `8388609.0f`), since adding `0.5f` rounds up to the next integer in those cases.  
The double version presumably has the same issues for slightly different values.

Test code:
```java
public class Test {
    public static int round(float a) {
        return a > 0F ? (int)(a + .5F) : (int)(a - .5F);
    }

    public static void main(String[] args) {
        int i = Integer.MIN_VALUE;
        do {
            final float f = Float.intBitsToFloat(i);
            if (round(f) != Math.round(f)) {
                System.out.println("Different for "+f+" (bits: "+Integer.toString(i)+")");
            }
            ++i;
        } while (i != Integer.MIN_VALUE);
    }
}
```
(Off-topic: Is there a better way to get a loop over all `int`s?)

FxMorin:
@malte0811 Good to know, guess it will be a carpet-fixes rules then

PR #340, opened 2022-01-20
add an entrypoint in LithiumMixinPlugin postApply

SFort:
- adds an interface `LithiumMixin` to the api with a default method `onLithiumMixinPostApply`
- calls a `lithium-mixin-post-apply` entrypoint in `LithiumMixinPlugin;postApply` which passes it's parameters using said interface

My usecase being fixing unascribed/fabrication#376 which requires the ability to toggle lithiums redstone optimizations at runtime sfort/fabrication@03c55af


SFort:
never mind, looks like i was being dumb and this is achivable without an entrypoint, sorry to  bother

PR #377, opened 2022-07-21
change: optimize EntityTrackingSection.isEmpty() a tiny bit

ishland:
This PR contains a micro optimization that brings MSPT down from 237 to 224 with 100 carpet fake players online after using `/spreadplayers` with a radius of 1024
Server simulation distance set to 8 and view distance set to 12


2No2Name:
This is now merged in a modified form, since mixin-extras provides a useful way around the overwrite

PR #392, opened 2022-09-13
Fixed crashes caused by LandPathNodeMakerMixin, changed path node type caching to load the node types when they are requiered and not in the start.

ghost:
This pull request does two things:

1) Removes the `@Overwrite` mixins in LandPathNodeMakerMixin, and replaces them with `@Inject`, that is more compatible with other mods and avoid crashes with some mods.
2) Adds pre-support (incomplete for now) for caching custom node types from fabric path node types api with lithium caching system ([PR #2519](https://github.com/FabricMC/fabric/pull/2519)).


devpelux:
I've changed the title, this means that this pr does not adds the complete support for the land path node types feature from fabric api, but fixes only the crash caused by the mixin.
Currently, due to the lower priority of fabric api, lithium overwrites what fabric does.
Needs another step to add full support, but this cannot be done by now because land path node types feature needs to be merged.

2No2Name:
This PR is now obsolete due to the mod compatibility changes in https://github.com/CaffeineMC/lithium-fabric/commit/254ace142d0abeaa2ed692081d3778e633acadcf

PR #437, opened 2023-01-03
fix: Update dependencies and fix building

burgerindividual:
This also fixes a bug with mixin-config-plugin that was windows specific


2No2Name:
Partially merged, the other changes in this PR seem to not work

2No2Name:
If there is more issues on windows, please comment again

PR #471, opened 2023-04-21
feat: optimized preloading chunks by removing streams

Crec0:
Overwrite chunk preloading method to not use streams. This gives a considerable server performance when large amount of entities are going through nether portals

### Theoretical test
Average mspt with with 1024 repeating command blocks spawning items that go through portals to other dimension

<table>
<tr>
 <td></td>
 <td>Overworld -> Nether</td>
 <td>Nether -> Overworld</td>
</tr>
<tr>
 <td>Without</td>
 <td>76</td>
 <td>425</td>
</tr>
<tr>
 <td>With</td>
 <td>62</td>
 <td>147</td>
</tr>
</table>

### Practical test
A stacking raid farm at 32mspt without optimization decreased to 24mspt with this optimization


2No2Name:
Looks good to me. I might modify this to include a conceptual change:
Instead of checking whether all sourrounding chunks were already preloaded, it might be faster to store that all necessary sourrounding chunks are preloaded for a given center chunk. However, this requires additional profiling to confirm that a significant amount of time is spent inside this function after your changes.

2No2Name:
I noticed that you are excluding the top most chunk section of every chunk. I assume this is unintentional and change it, unless you give a reason for it.

2No2Name:
Merged a modified version manually

PR #474, opened 2023-05-24
change: use main worker executor for util.block_tracking

embeddedt:
### Proposed Changes

Changes the `CompletableFuture` used in `mixin.util.block_tracking` to run on the normal Mojang background executor instead of the common ForkJoinPool. Use of the latter can cause issues with mods as it does not have the correct context classloader set (due to being initialized before the modloader). An example of such an issue is https://github.com/Fuzss/puzzleslib/issues/41; there were many mods involved but the problem originated with Lithium running code on the common ForkJoinPool in the first place.


2No2Name:
Thank you for this PR. Merged manually

PR #495, opened 2023-11-04
[WIP] Optimize item entity merging for full stacks

Andrews54757:
# What it does
Optimizes itemEntity `tryMerge` by only iterating over stacks that have `canMerge` true. Will store mergeable item entities in separate list, and remove them from the list when `canMerge` is no longer true.

## Results
Without optimization, using HermitCraft boxes from Storage Tech.

No optimization

With optimization
With optimization

## Future work

- [x] Clean up implementation


Andrews54757:
https://github.com/Andrews54757/lithium-fabric/pull/1
Further optimizations for stacks that are more than half full.

Andrews54757:
Closing due to latest commit

PR #496, opened 2023-11-09
Use order index instead of indexOf for item merging optimization

Andrews54757:
# What it does
Uses a pre-stored order index to insert item entities into the right group at the right location. O(n log n) -> O(log n). Important for keeping hopper operations optimized.

## Notes
- Order index is only set upon insertion into `SizeBucketedItemEntityList`
- I know @2No2Name doesn't like storing a seperate index because of issues in keeping it up to date, but this index is only updated once on insertion and improves time complexity for an important component.
- The order index is stored as a `long`
- I've noticed that you duplicated some code in `handleItemEntityStackReplacement`, since it looked like a mistake, I removed it.


Andrews54757:
closing due to latest commit

PR #497, opened 2023-11-09
fix: throw EntryMissingException for null entries in hash palette

embeddedt:
At some point (I believe it was 1.17 or 1.18), vanilla changed the default palettes used by PalettedContainers (including `BiMapPalette`) to throw `EntryMissingException` if the entry for the desired ID is null. This PR adjusts Lithium to do the same.


Nolij:
This looks like a fairly straightforward change. Any reason why it hasn't been merged yet?

2No2Name:
It hasn't been merged yet, because I have been quite busy

PR #501, opened 2023-12-18
Optimize FarmlandBlock nearby water lookup

EinS4ckZwiebeln:
Some neat farmland water lookup optimizations I stumbled upon in an old paper patch and ported over to fabric. I don't see any disadvantages in adding them since, to my knowledge and tests, this shouldn't break any vanilla behavior.

If there are any (style-) issues with this PR, please let me know since this is my first GitHub PR related to anything Java :)

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/block/farmland/FarmlandBlockMixin.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/block/farmland/FarmlandBlockMixin.java
@@ -1,0 +1,33 @@
+package me.jellysquid.mods.lithium.mixin.block.farmland;
+
+import net.minecraft.block.FarmlandBlock;
+import net.minecraft.fluid.FluidState;
+import net.minecraft.registry.tag.FluidTags;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.WorldView;
+import net.minecraft.world.chunk.Chunk;
+import org.spongepowered.asm.mixin.Mixin;
+
+@Mixin(FarmlandBlock.class)
+public abstract class FarmlandBlockMixin {
+
+    // Optimize FarmlandBlock nearby water lookup
+    private static boolean isWaterNearby(WorldView world, BlockPos pos) {
+        int posX = pos.getX();
+        int posY = pos.getY();
+        int posZ = pos.getZ();
+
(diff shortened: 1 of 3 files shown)


maityyy:
Please clarify in the comments to the code how exactly the optimization works, and also whether this is allowed under Paper license?

EinS4ckZwiebeln:
> Please clarify in the comments to the code how exactly the optimization works, and also whether this is allowed under Paper license?

License-wise, I don't really see why this shouldn't be allowed since Paper is under the GNU General Public License, which allows distribution and modification. 

Regarding the optimizations: Apparently the abstract block iteration was taking a lot of time.

You might as well take a look at the patch yourself: https://github.com/PaperMC/Paper/blob/e3140fb70e818b9e3fd7becbb9842e7f24847bb5/patches/server/0682-Optimise-BlockSoil-nearby-water-lookup.patch#L4

jellysquid3:
Paper is licensed under the GPLv3, which is stronger than the LGPLv3 (Lesser GPL) that Lithium uses. We can't weaken the GPL license, so it will not be possible to incorporate this patch unless the original author gives us a license grant under the LGPLv3.

EinS4ckZwiebeln:
> Paper is licensed under the GPLv3, which is stronger than the LGPLv3 (Lesser GPL) that Lithium uses. We can't weaken the GPL license, so it will not be possible to incorporate this patch unless the original author gives us a license grant under the LGPLv3.

Aight, didn't know that. Never mind then :/

jellysquid3:
To be entirely honest with you, I don't think the original patch is actually copyrightable... it is too trivial, and more or less direct translation of existing code, of which there is no other way to express it.

So I might argue that inclusion in Lithium is fine if only because it wouldn't pass other conditions for being copyrightable. However, that isn't a very popular stance to take, and it tends to make other people upset.

0-x-2-2:
Another thing to note about Paper is that some patches/code are licensed under MIT but unfortunately, this patch was published slightly after Spottedleaf https://github.com/PaperMC/Paper/commit/d80e43647dea345629d6ba965db798392e0ee997 removed their wishes of future patches to be licensed under MIT. That said this patch is pretty trivial so you could most likely just ask Spottedleaf for permission to use it under MIT or LGPLv3.

PR #509, opened 2024-04-09
Fix hashed reference lists being created unnecessarily

embeddedt:
The original logic creates a `HashedReferenceList` when the list is very small (which is when the optimization would have least impact). This generates a lot of allocations when creating tiny `Pool`s and also seems unintended. Now, the hashed implementation is used for large Pools (where it is likely to improve performance), and the list is used as-is if small enough.


embeddedt:
Already fixed by https://github.com/CaffeineMC/lithium-fabric/commit/be8ee4d9e48c69bd1df47cd541fc7d7e06693836

PR #565, opened 2024-09-13
Fix incompatability with scarpet explosion event

wendavid552:
Since https://github.com/gnembon/fabric-carpet/pull/1937 didn't get merged for a while, here is a tricy ways of fixing the incompatability in lithium's side.

Just simply replace the `@Redirect` into `@WrapOperation`, and the `original.call()` would contain the `@Redirect` from fabric-carpet. Therefore the `getExplosure()` firstly goes to our wrapping, then scarpet would detect the explosion event. But if it is a duplucated explosure we would skip it, and actually from the design of scarpet it is useless to detect the explosion again (as they are the same one, just mentioned again).

Tested in 1.21 with `fabric-carpet`, the wrapping works and the explosion event is triggered too.


2No2Name:
Thank you for this contribution. I am a bit concerned about the performance impact of `@WrapOperation` but many users suggested that the compiler should be able to eliminate it. Performance testing will be done in the future. Contributions are also welcome

PR #567, opened 2024-10-04
Remove the `SINE_TABLE` array in `MathHelper`

### LINKED ISSUE #381: Remove the `SINE_TABLE` array in `MathHelper`
Lithium replaces the `sin` and `cos` methods in `MathHelper`, the only two places that access its `SINE_TABLE` array. That means that the array is unused after that, however it is still kept in memory.

Suggestion to remove it or at least make it an empty array.

Colonel-Sandvich:
Closes #381

### DIFF
--- a/src/main/java/me/jellysquid/mods/lithium/mixin/math/sine_lut/MixinMathHelper.java
+++ b/src/main/java/me/jellysquid/mods/lithium/mixin/math/sine_lut/MixinMathHelper.java
@@ -1,7 +1,6 @@
 package me.jellysquid.mods.lithium.mixin.math.sine_lut;
 
 import me.jellysquid.mods.lithium.common.util.math.CompactSineLUT;
-import net.fabricmc.fabric.api.event.lifecycle.v1.ServerLifecycleEvents;
 import net.minecraft.util.math.MathHelper;
 
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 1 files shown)


src/main/java/me/jellysquid/mods/lithium/mixin/math/sine_lut/MixinMathHelper.java
```diff
@@ -1,12 +1,35 @@
...
+    public static float[] SINE_TABLE;
+
+    @Inject(method = "<clinit>", at = @At("RETURN"))
+    private static void onClassInit(CallbackInfo ci) {
+        new CompactSineLUT(); // Force class initialisation
+        MixinMathHelper.SINE_TABLE = null;
+    }
+
+    static {
+        ServerLifecycleEvents.SERVER_STARTING.register((server) -> {
```
2No2Name: The above injection should be sufficient already, I don't think this part is needed

2No2Name (commented):
PR is fine, the concerns already mentioned in the issue are mod compatibility. I don't expect this change to have any measurable effect on performance.

src/main/java/me/jellysquid/mods/lithium/mixin/math/sine_lut/MixinMathHelper.java
Colonel-Sandvich: Yeah I'm new to Minecraft modding so I added both methods because I was unsure which one would be preferred.

2No2Name (approved):

src/main/java/me/jellysquid/mods/lithium/mixin/math/sine_lut/MixinMathHelper.java
```diff
@@ -2,11 +2,27 @@
...
 
 @Mixin(MathHelper.class)
 public class MixinMathHelper {
+    @Shadow
+    @Mutable
+    private static float[] SINE_TABLE;
+
+    @Inject(method = "<clinit>", at = @At("RETURN"))
+    private static void onClassInit(CallbackInfo ci) {
+        new CompactSineLUT(); // Force class initialisation
```
Earthcomputer, line 22: A better way to do this is to make an empty public static method and call it

PR #568, opened 2024-10-07
Fix crash with betterchunkloading mod

lugui1998:
There is a issue involving Betterchunkloading mod that causes the server to crash when a player moves enough to unload chunks.
``this.positionWithExpiringTicket.get(pos)`` are presumed to never be null but rather a list that could be empty. More debugging is needed in order to find the reason why this would ever return a null value, which is probably Betterchunkloading's fault, regardless, this check will prevent the server from crashing on a situation like this.


2No2Name:
I interpret the given NullPointerException issue at a hint at incompatible changes in the two mods. Adding a simple null check may completely resolve the issue, but it is also very likely that there are other incompatibilities which are hidden by the additional null check. I think investigating why the NullPointerException happens in the first place is needed.

PR #583, opened 2024-11-11
Update redstone wire optimizations and fix #581

### LINKED ISSUE #581: Experimental Redstone Features: Indefinitly powered redstone dust
### Version Information

> lithium-fabric-0.14.1-snapshot+mc1.21.3-build.89

### Expected Behavior

> Redstone dust power set to 0 if power source is removed.

### Actual Behavior

> Any redstone line with a length of at least two blocks stays powered, even after the redstone source is removed

### Reproduction Steps

> 1. Create a world with "Redstone Experiments" enabled
> 2. Place two redstone dust next to each other
> 3. Place a redstone block to power the redstone dust
> 4. Break the redstone block
> 5. The redstone dust stays powered

### Other Information

> Tested with no aditional mods, the bug still occurs.

SpaceWalkerRS:


common/src/main/java/net/caffeinemc/mods/lithium/common/block/redstone/RedstoneWirePowerCalculations.java
```diff
@@ -0,0 +1,222 @@
...
+    private static final int MAX = 15;           // largest possible power value
+    private static final int MAX_WIRE = MAX - 1; // largest possible power a wire can receive from another wire
+
+    private static Block wireBlock;
+    private static RedstoneWireEvaluator evaluator;
+    private static boolean ignoreWires = false;
+    private static boolean ignoreNonWires = false;
+
+    public static int getNeighborBlockSignal(Block wireBlock, RedstoneWireEvaluator evaluator, Level level, BlockPos pos) {
+        ignoreWires = true;
```
2No2Name, line 68: I am concerned about these static fields, mod compatibility with threading mods will definitely suffer

2No2Name: I think converting them to a local variable should be possible if I am not misunderstanding the code

2No2Name (approved):
Since you are the original author of this optimization and also the tests are passing, I mostly trust that the code is correct. When merging I will change the static fields to local variables, as I do not see a good reason to keep them static.

common/src/main/java/net/caffeinemc/mods/lithium/common/block/redstone/RedstoneWirePowerCalculations.java
SpaceWalkerRS: yeah that'd work

2No2Name:
Rebased onto develop and replaced the static fields with local variables.

Thanks for the PR

PR #507, opened 2024-03-16
Optimized explosion entity exposure calculations

crosby-moe:
## Overview  
This pull request comprises three optimizations:
- Optimize exposure raycasts by removing fluid handling, getting hit face, and other useless computations.
- Short-circuit exposure calculations on discarded entities.
- Remove duplicate `Explosion#getExposure` calls.

Since these optimizations specifically target entity-related explosion calculations, they have virtually no effect on single explosions. Chained explosions, however, see massive benefits due to the creation of dropped items & exp orbs which are then affected by the subsequent explosions.

I'm not particularly good at documentation, so suggestions are welcome.

## Benchmarks
~150% improvement in time spent exploding 4096 tnts.
Benchmarked using seed `-3199243952349285632` (randomly chosen) and 4096 tnt (`/fill ~1 ~-7 ~ ~16 ~8 ~15 tnt`)

Before:
9492 ms spent in `World.createExplosion()` [Spark profile](https://spark.lucko.me/f0gKvv0dY6)
before

After:
3832 ms spent in `World.createExplosion()` [Spark profile](https://spark.lucko.me/DzkUu3WLZ4)
after

PR #623, opened 2025-02-23
Optimize explosion entity raycasts

crosby-moe:
## Overview

This pull request does two things:
1. Fixes the [unused `HashSet` allocation redirect](https://github.com/CaffeineMC/lithium/blob/947217c2908f81c97e864a85f1dface7b00fac2e/common/src/main/java/net/caffeinemc/mods/lithium/mixin/world/explosions/block_raycast/ServerExplosionMixin.java#L106-L112) in `block_raycast.ServerExplosionMixin` that was removed during the 1.21.3 update of lithium (`NEW` redirects *have* to return a non-null object.)
2. Readds and improves upon the specialized raycast hit factory that I had added in my earlier pull request (#507)

## Benchmarks
~80% improvement in time spent exploding 4096 tnts.
Benchmarked using seed `-3199243952349285632` (randomly chosen) and 4096 tnt (`/fill ~1 ~-7 ~ ~16 ~8 ~15 tnt`)

Before:
6612 ms spent in `World#createExplosion` (5102 ms spent in `ExplosionImpl#damageEntities`)
image

After:
3710 ms spent in `World#createExplosion` (1982 ms spent in `ExplosionImpl#damageEntities`)
image

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/lithium/common/world/explosions/MutableExplosionClipContext.java
+++ b/common/src/main/java/net/caffeinemc/mods/lithium/common/world/explosions/MutableExplosionClipContext.java
@@ -1,0 +1,15 @@
+package net.caffeinemc.mods.lithium.common.world.explosions;
+
+import net.minecraft.world.level.Level;
+import net.minecraft.world.phys.Vec3;
+
+public class MutableExplosionClipContext {
+    public final Level level;
+    public final Vec3 to;
+    public Vec3 from = null;
+
+    public MutableExplosionClipContext(Level level, Vec3 to) {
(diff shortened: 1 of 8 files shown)


2No2Name:
The `BlockGetter.clipWithInteractionOverride` is replaced with `VoxelShape.clip` because the raycast result's direction on the block does not matter for explosions

2No2Name:
This PR causes https://github.com/CaffeineMC/lithium/issues/630 due to removing the EntityCollisionContext, which affects raycasts, e.g. through scaffolding blocks. This is fixed in 0.15.1

PR #677, opened 2025-08-31
Update ServerExplosionMixin.java

Mr-Squid42:
i don't know if this would have any substantial performance benefits but it makes more sense to exit earlier in the function if the block is air and !this.explodeAirBlocks

it just seems a waste to me to calculate blast resistance of an air block as it remains constant


2No2Name:
```
        public static void main(String[] args) {
            System.out.println("0.3f * 0.3f == 0.9f is: " + (0.3f * 0.3f == 0.9f));
        }
```

2No2Name:
0.3f * 0.3f == 0.09f
This is correct but your suggestsions don't really improve anything relevant, unless you provide measurements, which really won't be possible
For your future pull requests please include measurements whether the performance increased ingame

2No2Name:
Also I am a bit concerned about mods which modify the effect of air blocks on the explosion rays. Lithium is intended to not modify those observable behaviors, even in minor details

PR #682, opened 2025-10-05
Improved Portal Search

jcw780:
Rewriting parts of NearbyPointOfInterestStream to remove the chunk sorting at the start and prioritize closer subchunks first - current one prioritizes closer chunks.
Some other improvements aim at returning POIs early were also included. 

Tradeoff is that POI subchunks need to get sorted so the worst case is worse - if a lot of subchunks get queued and sorted - although this should not be a common occurrence.

Performance:
See discord

Not entirely satisfied with the worst relative case results but it likely is better in more scenarios than the current implementation

Note: 
This is still a draft as of me opening this - this is why the new NearbyPointOfInterestStream in a 2nd file.
I have tested it for non-vanilla linking but it is possible that something was overlooked.


2No2Name:
Merged manually

PR #702, opened 2025-12-13
Chunk-Aware Supporting Block Search

jcw780:
Use ChunkAwareBlockCollisionSweeper for entity supporting block check.
Profiles in discord.


2No2Name:
Merged manually

PR #713, opened 2026-02-02
Fix ensureLoadedAndValid overwrite

jcw780:
Fix Lithium's ensureLoadedAndValid overwrite

Correct chunk loading conditions and loading order of ensureLoadedAndValid.
Vanilla loads when chunk has empty sections or invalid PoiSections - current implementation is the opposite (likely missed the !)
Vanilla checks the sections in x then y then z - current Lithium implementation checks y, z, then x.
This results in chunk loading order differences to vanilla. 
While it is unclear that anything is affected by this, I am opting to fix this just in case.


2No2Name:
Rebased and squashed

PR #718, opened 2026-02-21
Nitwit job search

jcw780:
- Remove jobsite AcquirePoi behavior if the villager only looks for PoiType.None (nitwits)
  - Profiles in discord
- Added brain mixin to make removing behaviors easier in the future


common/src/main/java/net/caffeinemc/mods/lithium/mixin/ai/useless_behaviors/BrainMixin.java
```diff
@@ -0,0 +1,60 @@
...
+    private Map<Activity, Set<Pair<MemoryModuleType<?>, MemoryStatus>>> activityRequirements;
+
+    @Shadow
+    @Final
+    private Map<Activity, Set<MemoryModuleType<?>>> activityMemoriesToEraseWhenStopped;
+
+    /**
+     * @author jcw780
+     * @reason This is so one can just null out behaviors earlier to prevent them from being added - otherwise the game will crash.
+     * Note: Could also use a sentinel instead of null if preferred.
```
2No2Name: Using a sentinel value might be better, in case some other mod breaks the null filtering mixin or uses them in between. The sentinel could be a behavior that never executes, has no internal state and is stored as a constant in some static field.

common/src/main/java/net/caffeinemc/mods/lithium/mixin/ai/useless_behaviors/BrainMixin.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Overwrite
```
2No2Name: If we can avoid using `@Overwrite` that would be great. If it is not possible here, it's also okay.

jcw780: fixed in recent commit

common/src/main/java/net/caffeinemc/mods/lithium/mixin/ai/useless_behaviors/BrainMixin.java
jcw780: should be fixed in recent commit

2No2Name:
Squashed to

https://github.com/CaffeineMC/lithium/commit/4af2386e9b825d2ea97dcc737520bba30af75a3a

and minor changes in

https://github.com/CaffeineMC/lithium/commit/2886705c42b75621398a3b949bbd5f785f65031b

PR #755, opened 2026-07-04
Fix LongJumpChoiceList returning wrong jump weights in get/remove and crashing when picking from an empty list

gurrveerrrr:
Offsets at array index i represent a squared distance of i + 1, but get(int) and remove(int) incorrectly used i as the distance. This caused get to return weights that were one distance too low and remove to corrupt weightByDistanceSq and totalWeight, eventually causing removeRandomWeightedByDistanceSq to crash or isEmpty() to disagree with size().

Also return null instead of calling Random.nextInt(0) when the list is empty, matching vanilla WeightedRandom.getRandomItem, which returns Optional.empty().


2No2Name:
Thank you for hunting down the issue and creating the PR! Have you encountered this issue in the game?

gurrveerrrr:
> Thank you for hunting down the issue and creating the PR! Have you encountered this issue in the game?

Thanks! I actually found it while looking through the code. I really like the mod and wanted to contribute over summer break, so I started reading through parts of the code and found the bug

PR #759, opened 2026-07-09
Fix multipart entities (dragon parts, modded part entities) missing from movement collision checks, matching vanilla Level.getEntities (fixes #651)

### LINKED ISSUE #651: Iron's Spells 'n Spellbooks incompatibility, shields have no collision
### Version Information

lithium-neoforge-0.15.0+mc1.21.1

### Expected Behavior

Shields from shield spell have collisions and stop entities

### Actual Behavior

Entities can pass through shields freely

### Reproduction Steps

1. Cast shield
2. Lure mob through it
3. It goes through it

### Other Mods

1. Iron's Spells 'n Spellbooks [1.21.1-3.12.0]
2. Curios API [9.5.1+1.21.1]
3. Geckolib [4.7.5.1]
4. playerAnimator [2.0.1]

gurrveerrrr:
While looking into entity collisions, I found that Lithium's optimized movement collision path wasn't checking multipart entity pieces like ender dragon parts.

Vanilla's `Level.getEntities` always includes these parts, but Lithium's movement collision code only queried entities stored in the section storage. Lithium already has a lookup that includes multipart entities (`getEntitiesOfEntityGroupPlusDragonPieces`), but it was only being used for projectile collisions.

This meant that collidable multipart entities from mods, such as the shields in Iron's Spells 'n Spellbooks (see #651), had no collision when Lithium's entity collision optimization was enabled because their part entities were never returned.

This change updates both movement collision paths to use the multipart-aware lookup and makes `entityFilter` nullable, since movement collisions don't pass one anyway. Vanilla behavior is unchanged, since vanilla dragon parts don't override `canBeCollidedWith`.

Fixes #651

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/lithium/common/services/PlatformEntityAccess.java
+++ b/common/src/main/java/net/caffeinemc/mods/lithium/common/services/PlatformEntityAccess.java
@@ -3,6 +3,7 @@ package net.caffeinemc.mods.lithium.common.services;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.phys.AABB;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.function.Predicate;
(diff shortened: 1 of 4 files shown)


2No2Name:
LGTM, however, I added an additional fix for the missing isSpectator check on Dragon Pieces / PartEntity

PR #760, opened 2026-07-10
Optimize jigsaw free space tracking with cuboid difference shapes

### LINKED ISSUE #235: Optimize VoxelShapes.combine for jigsaw structures
So I was stress testing my jigsaw structures by spawning a ton of structures and I noticed that while everything works great, the chunk loading is significantly slowed. So I decided to see what the bottleneck is and it turns out, it's pretty much nearly entirely just VoxelShapes.combine. In the profiler pictur ebelow you will see I was using my own StructurePoolBasedGenerator class which is a copy of vanilla's with some changes but even vanilla suffers from the same issue if you make every chunk spawn a larger village. The vanilla code is at StructurePoolBasedGenerator$StructurePoolGenerator line 244
```java
mutableObject4.setValue(
  VoxelShapes.combine(
     (VoxelShape)mutableObject4.getValue(), 
     VoxelShapes.cuboid(Box.from(blockBox4)), 
     BooleanBiFunction.ONLY_FIRST
   )
);
```

The profiler to show that VoxelShapes.combine takes up the majority of worldgen time over everything else. (This is with Lithium on so even Lithium's current optimizations to VoxelShapes class isn't helping much)
image

What is happening at that line is the jigsaw structure piece is now determined that it will spawn. And so it takes the structure's current VoxelShape of its boundaries and merges the new piece's bounding box into that VoxelShape to create a larger and more complex shape that encompasses the structure pieces in a tight fit. Now with jigsaw structures like bastions or villages, there can be a huge number of individual pieces that has to be merged into the overall structure VoxelShape. And custom structures can get even larger and have even more branches which would explode the VoxelShapes.combine method even more. The more pieces a structure has, the worse the issue becomes I believe.

That's why I think this could be an interesting target for Lithium to optimize. Maybe by redirecting just that VoxelShapes.combine method for jigsaw structures, a better method could be used that is solely optimized for adding new bounding boxes to a complex shape? Just an idea. 

I tried to look into how VoxelShapes.combine works and my brain imploded lol

gurrveerrrr:
Fixes #235

Jigsaw structure generation spends a lot of time repeatedly combining `VoxelShape` while checking and subtracting free space. This replaces that hot path with a lightweight cuboid-difference representation that avoids building voxel shapes for the common case.

The fast path handles vanilla jigsaw placement exactly while falling back to the original implementation for anything it can't represent (such as off-grid or modded shapes), preserving vanilla behavior.

Added randomized tests that compare the new implementation against vanilla across full placement chains, boundary cases, shared-list divergence, and fallback paths.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/lithium/common/shapes/VoxelShapeCuboidDifference.java
+++ b/common/src/main/java/net/caffeinemc/mods/lithium/common/shapes/VoxelShapeCuboidDifference.java
@@ -1,0 +1,495 @@
+package net.caffeinemc.mods.lithium.common.shapes;
+
+import com.mojang.math.OctahedralGroup;
+import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
+import it.unimi.dsi.fastutil.doubles.DoubleList;
+import net.minecraft.core.AxisCycle;
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
+import net.minecraft.world.phys.AABB;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.Vec3;
+import net.minecraft.world.phys.shapes.BooleanOp;
+import net.minecraft.world.phys.shapes.DiscreteVoxelShape;
+import net.minecraft.world.phys.shapes.Shapes;
+import net.minecraft.world.phys.shapes.VoxelShape;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * A {@link VoxelShape} that represents a cuboid minus a list of cuboids without storing any voxels. It is used to
+ * replace the very expensive incremental {@link Shapes#joinUnoptimized} / {@link Shapes#joinIsNotEmpty} calls jigsaw
+ * structure placement uses to track the remaining free space (structure bounds minus already placed pieces).
+ * <p>
+ * Instead of merging coordinate lists and joining voxel sets, subtracting a placed piece just appends 6 doubles to a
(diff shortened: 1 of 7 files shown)


2No2Name:
Have you checked the performance impact of these changes?

2No2Name:
https://modrinth.com/mod/structure-layout-optimizer is a mod that optimizes this part of the code already I believe

gurrveerrrr:
> Have you checked the performance impact of these changes?

I haven't run proper benchmarks yet. I've mostly been focused on making sure it matches vanilla first, and all the randomized tests have been passing. Im right now running a microbenchmark that runs realistic jigsaw placement chains through both the vanilla Shapes.join path and the new cuboid-difference path and i'll share the results.

gurrveerrrr:
> https://modrinth.com/mod/structure-layout-optimizer is a mod that optimizes this part of the code already I believe

I think it's doing something a bit different. It looks like it reduces how many pieces get checked, while this changes how the free space is tracked

gurrveerrrr:
Did some benchmarks with the same operation order that `tryPlacingChildren` does (free space join, reject check, subtract) using same inputs for both paths.

results:

small: 4.1ms -> 0.33ms (~13x)
medium: 146ms -> 0.26ms (~560x)
village size: 2355ms -> 0.33ms (~7200x)

the reason the gap gets so big is vanilla keeps re checking the growing voxel grid every time, while the new representation stays pretty much constant.

also found a issue while testing. calling `VoxelShape` methods on the new shape makes it materialize back to vanilla, which made my first results wrong. these numbers avoid that and match how jigsaw actually uses it.

this only tests the shape operations, not the full structure generation time. stuff like template loading is still the same, but since #235 showed these joins are a big bottleneck this should still help a lot.

also this should work alongside structure-layout-optimizer. that reduces how many checks happen, while this makes the checks cheaper.

PR #768, opened 2026-07-29
Fix target block validation logic in RemoveBlockGoalMixin

Spagles:
Correct the logic for validating target blocks in the `RemoveBlockGoalMixin` to ensure proper functionality when checking for air blocks above the specified position. This change improves the accuracy of block removal goals.


2No2Name:
Whether we call `move(0,1,0)` twice or `above(1)` followed by `above(2)` doesn't really matter, but the previous version actually saves on the allocation of the new block position objects. I don't believe this PR's change is an improvement

Issue #772, opened 2026-08-22
Client kicked with "Network Protocol Error" when a sleeping villager dies (experimental unused_brain mixin)


RasmusKD:
### Version

lithium-fabric-0.24.2+mc26.1.2 (Fabric, singleplayer, Windows). Shipped in the Simply Optimized Reloaded modpack, which sets `mixin.experimental=true`.

### Reproduction Steps

1. Enable Lithium's experimental mixins (`mixin.experimental=true` in `lithium.properties`), so `mixin.experimental.client_tick.entity.unused_brain` is active.
2. Kill a **sleeping** villager (for example TNT next to a house at night).
3. The client is kicked from the (integrated) server with **"Network Protocol Error"**.

### Analysis

The death entity event reaches the client, and vanilla's client-side death path writes to the villager's brain:

`ClientboundEntityEventPacket` → `Villager.handleEntityEvent` → `LivingEntity.die` → `Villager.stopSleeping` → `Brain.setMemory`

`stopSleeping` runs on both sides in vanilla, so the dummy client-side brain introduced by `mixin.experimental.client_tick.entity.unused_brain` throws its guard exception inside packet handling, which disconnects the client. Any mob whose client-side vanilla code touches brain memories on death while sleeping should trigger the same guard.

### Crash / Stacktrace

```
[Render thread/ERROR]: Failed to handle packet net.minecraft.network.protocol.game.ClientboundEntityEventPacket@6532fe94, disconnecting
java.lang.UnsupportedOperationException: Dummy client side brain memory slot cannot be modified! This is an optimization introduced by lithium. [...]
	at net.minecraft.world.entity.ai.memory.MemorySlot.throwOnModifyDummyMemorySlot(MemorySlot.java:1032)
	at net.minecraft.world.entity.ai.memory.MemorySlot.handler$dbi000$lithium$throwIfDummyMemorySlot(MemorySlot.java:1026)
	at net.minecraft.world.entity.ai.memory.MemorySlot.set(MemorySlot.java:38)
	at net.minecraft.world.entity.ai.Brain.wrapOperation$cml000$lithium$setTrackingChanges(Brain.java:1157)
	at net.minecraft.world.entity.ai.Brain.setMemoryInternal(Brain.java:256)
	at net.minecraft.world.entity.ai.Brain.setMemory(Brain.java:202)
	at net.minecraft.world.entity.npc.villager.Villager.stopSleeping(Villager.java:949)
[...]

PR #773, opened 2026-08-22
Skip client-side brain write when a sleeping villager dies

RasmusKD:
Closes the residual from #772.

With `mixin.debug=true`, the dummy-brain tripwire still fires a false positive on a vanilla path: when a sleeping villager dies, the client runs `LivingEntity.die` -> `Villager.stopSleeping` -> `Brain.setMemory(LAST_WOKEN, ...)`, and the throw lands inside packet handling, disconnecting the client. In production the write is already cancelled silently, so this only affects debug users, but it is vanilla writing client-side rather than a misbehaving mod, which is what the tripwire is meant to catch.

This adds a `VillagerMixin` guarding the `stopSleeping` write with `@WrapWithCondition`, in the same shape as the existing `AllayMixin` whitelist for `mobInteract`.

Tested:
- `:fabric:build` succeeds. Two optional gametests failed on my machine (`ai_villager_hide_in_home` timeout, `lava_push_speed` timing); the gametest server has no client, so `level().isClientSide()` is always false there and the wrapped call is unconditionally original, which makes those look like pre-existing timing flakes rather than effects of this change.
- The production behavior around the same path was verified live on 26.2 with 0.25.3: killing sleeping villagers disconnects nothing and the entities die normally, unchanged by this patch.

### DIFF
--- a/common/src/main/java/net/caffeinemc/mods/lithium/mixin/client_tick/entity/unused_brain/VillagerMixin.java
+++ b/common/src/main/java/net/caffeinemc/mods/lithium/mixin/client_tick/entity/unused_brain/VillagerMixin.java
@@ -1,0 +1,34 @@
+package net.caffeinemc.mods.lithium.mixin.client_tick.entity.unused_brain;
+
+import com.llamalad7.mixinextras.injector.v2.WrapWithCondition;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EntityType;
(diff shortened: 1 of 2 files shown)


RasmusKD:
Makes sense - with vanilla itself legitimately writing client-side, per-call-site whitelisting is the wrong shape, and the real risk is a later read of data that was never written. Closing. If a read-side detection lands at some point I am happy to test it against the sleeping-villager path.

